PR #98, opened 2020-06-14
Feature nithin

maaverik:
Queue name validation to allow only - and alphanumeric characters and changed font for the description.


maaverik:
Error handling; linting and formatting rules

PR #104, opened 2020-06-23
Created automated linter workflow

maaverik:
Modeled after https://github.com/samuelmeuli/lint-action

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,29 @@
+name: Lint
+
+on: push
+
+jobs:
+  run-linters:
+    name: Run linters
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out Git repository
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,29 @@
+name: Lint
+
+on: push
```
daltonfury42, line 3: let's do on all PRs to master:
```
on:
  pull_request:
    branches: [master]
```

Then we can add a branch protection rule to allow merging only if the check passes

daltonfury42:
It's breaking though

maaverik:
From what I understand, it breaks cause it expects eslint and prettier to be present already, but that's not the case for this commit.

maaverik:
It still doesn't detect eslint and prettier. I thought it uses the same package.json as the project, but still fails saying eslint is not found. Have to do more research.

PR #107, opened 2020-06-26
Changed directory to run npm install in

maaverik:
Should fix the issue with lint action not working


maaverik:
Still fails even though npm install ran fine this time

daltonfury42:
This is how you can set the working directory https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaultsrun

daltonfury42:
I guess it's not because of directory. One thing we can do is to report the issue upstream..

For backend, the way @thehamzarocks has set it up is to automatically lint and fix errors. Is frontend also like that?

maaverik:
Could you merge this with your token service branch? It has outdated changes like apis.js and queue.js

PR #122, opened 2020-07-05
Cleaned up code

maaverik:


simplq/src/components/pages/Join/index.jsx
```diff
@@ -4,7 +4,7 @@ import JoinQueueForm from './Form';
 import { setTokenId, setJoinerStep, setAheadCount, setQueueId } from '../../../store/appSlice';
 import * as TokenService from '../../../services/token';
 import { handleApiErrors } from '../../ErrorHandler';
-import { SimplQHeader } from '../../common/Header.stories';
+import { SimplQHeader } from '../../common/Header/Header.stories';
```
daltonfury42, line 7: Here it should be just `import { SimplQHeader } from '../../common/Header'`

simplq/package.json
```diff
@@ -31,7 +31,7 @@
     "build": "react-scripts build",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "lint": "eslint src --ext .js,.jsx --fix-dry-run",
+    "lint": "eslint src --ext .js,.jsx --fix",
```
daltonfury42, line 34: Do we need these filters by extension? Otherwise, should we add json too?

same for the lint-staged one.

daltonfury42: Remove fix-dry-run from lintstaged too, line 79

.github/workflows/lint.yml
```diff
@@ -3,6 +3,11 @@ name: Lint
 on: 
   pull_request:
       branches: [master]
```
daltonfury42, line 5: add new-frontend and prod also to this

simplq/package.json
maaverik: It didn't run on all files when I tried using the glob syntax (src/**/**), that's why I used this.

maaverik:
Looks like the extra brace is caused by the removing the fix-dry-run somehow. Maybe adding json will help.

maaverik:
It might be a conflict between eslint and prettier during the fix. Could you look into this?

daltonfury42:
ok. lets merge this one.

PR #123, opened 2020-07-05
Eslint husky fixes

daltonfury42:
@maaverik The pre-commit hook part, (husky, lint-staged setup) was not working at my end. It' working after this change.


simplq/src/serviceWorker.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
daltonfury42, line 1: I've disabled linting in this file. Any idea what this file is about?

simplq/src/services/axios.js
```diff
@@ -1,9 +1,9 @@
-import  axios from 'axios';
+import axios from 'axios';
 import { getAccessToken } from './auth';
 
 const baseURL = 'https://devbackend.simplq.me/v1';
 
-export const makeRequest = async(method, url, data) => {
+export default async (method, url, data) => {
```
daltonfury42, line 6: Fix for `error  Prefer default export  import/prefer-default-export`

simplq/src/store/appSlice.js
```diff
@@ -14,49 +14,58 @@ const appSlice = createSlice({
   reducers: {
     setQueueName: (state, action) => {
       const queueName = action.payload;
+      // eslint-disable-next-line no-param-reassign
```
daltonfury42, line 17: I ended up disabling `no-param-reassign` for these lines. Looks like immerjs (the immutablility library redux toolkit uses behind the scenes) and this rule doesn't go well together, and this is the easiest fix I could come up with.

https://github.com/immerjs/immer/issues/189

simplq/amplify/team-provider-info.json
```diff
@@ -18,5 +18,3 @@
         }
       }
     }
-  }
```
daltonfury42, line 21: extra curley braces had somehow crept into this file and the above one

simplq/src/serviceWorker.js
maaverik: It's for enabling local caching on the final site. It's not enabled right now because it'll make development harder with changes not showing up on saves. We'll enable after the whole new UI is done. You don't need to disable linting though. Could you look at new-frontend? I'm pretty sure I fixed all the errors by adding global rules instead of inline rules when it makes sense.

maaverik:
I think most of these changes are unnecessary because they've been taken care of in new-frontend by global rules. We can just add all required fixes there and move to master together.

simplq/src/serviceWorker.js
daltonfury42: let me check

daltonfury42:
there is a misconfiguration somewhere... the husky setup is not working, or something. If I run npm run lint, it's giving this:

image

One thing we should do is, use `--fix` instead of `--fix-dry-run`, it will fix most issues automatically, taking the pain from our hands.

The extra braces are also there... we can discuss on call when you have time.

maaverik:
This is how I've designed to make it work. So husky doesn't directly use npm run lint, it uses lint-staged which makes it run the linter on only the staged files. I had fixed a lot of the common errors before starting on the new frontend and left the ones on files we might modify later on and hence wouldn't need to waste time fixing. In this case, we don't need CentralSection anymore, so the file can be removed, the error boundary doesn't work yet, so left that for now, and we're going to modify appSlice, so left that too. We could just fix linting on these files when we work on them. Essentially, at the moment, linting only comes into play on the files we modify (staged files).

daltonfury42:
Ok. That makes sense. I didn't see your other PR, saw that just now. Let's get that merged first.

PR #129, opened 2020-07-05
polling for status changes in admin and waiting pages

thehamzarocks:


simplq/src/components/pages/Admin/index.jsx
```diff
@@ -25,14 +25,20 @@ export default () => {
 
   const [items, setItems] = useState();
 
-  const update = () => {
+  const update = (updateClicked) => {
     if (queueId) {
       QueueService.get(queueId)
         .then((data) => {
           setItems(data.tokens);
+          if (!updateClicked) {
```
thehamzarocks, line 33: Using updateClicked to determine if this is a call from polling or from a button click might be messy. Shall I just create a separate function pollForUpdates instead of using if's everywhere?

simplq/src/components/pages/QueueStatus.jsx
```diff
@@ -64,10 +77,6 @@ function QueueStatus() {
       });
   };
 
-  if (aheadCount == null && !updateInProgress) {
```
thehamzarocks, line 67: I presume we don't need this anymore

simplq/src/components/pages/Admin/index.jsx
```diff
@@ -62,7 +68,7 @@ export default () => {
         queueId={queueId}
         className={styles.shareButton}
         onRefresh={() => {
-          update();
+          update(true);
           setItems(false);
```
thehamzarocks, line 72: This seems redundant. Why was this added? Can it be removed safely?

thehamzarocks: I believe it can

daltonfury42: Do you want to completely rewrite this page? not the HTML/Styling part, but the logic?

thehamzarocks: Hmm ok. I'll take it up. And the queustatus file too

simplq/src/components/pages/Admin/index.jsx
maaverik: Yes, please. It'll be cleaner.

simplq/src/components/pages/Admin/index.jsx
```diff
@@ -37,6 +37,22 @@ export default () => {
...
+        .then((data) => {
+          setItems(data.tokens);
+          setTimeout(pollAndUpdate, 10000);
+        })
+        .catch((err) => {
+          handleApiErrors(err);
+          setTimeout(pollAndUpdate, 10000);
+        });
+    }
+  };
```
daltonfury42, line 52: This is a very interesting problem. I think we can have a single function that takes care of both. Then the function can be called at start using `useEffect` at start and also when the button is clicked.

Use `clearTimeout` to cancel the previous timeout when the function is called again. Can you give this a try? The thing to figure out is where to store the `timeoutID`

simplq/src/components/pages/QueueStatus.jsx
daltonfury42: shitty code.. 😓

simplq/src/components/pages/Admin/index.jsx
thehamzarocks: A recommended way involves using a higher-order component called timeout which has timeout methods and passing in those methods as props to our child component.

An alternative way would be to use the component's state.

simplq/src/components/pages/Admin/index.jsx
```diff
@@ -24,19 +24,26 @@ export default () => {
...
   const [items, setItems] = useState();
+  const [timeoutId, setTimeoutId] = useState();
 
   const update = () => {
+    clearTimeout(timeoutId);
     if (queueId) {
       QueueService.get(queueId)
         .then((data) => {
           setItems(data.tokens);
+          const timeout = setTimeout(update, 10000);
```
daltonfury42, line 35: declare this as const on top: 10000

maaverik:
If someone approves, all subsequent commits on the PR don't need approval to merge. This is not how it should be, right? Is there some setting for this?

simplq/src/components/pages/Admin/index.jsx
```diff
@@ -26,17 +38,24 @@ export default () => {
   const [items, setItems] = useState();
 
   const update = () => {
+    clearTimeout(timeoutUtil.getTimeoutId());
     if (queueId) {
       QueueService.get(queueId)
         .then((data) => {
           setItems(data.tokens);
+          const timeoutId = setTimeout(update, 10000);
```
thehamzarocks, line 46: could make it a single-liner

simplq/src/components/pages/QueueStatus.jsx
```diff
@@ -37,21 +49,25 @@ function QueueStatus() {
   const classes = useStyles();
 
   const update = () => {
+    clearTimeout(timeoutUtil.getTimeoutId());
     if (tokenId) {
-      setUpdateInProgress(true);
```
thehamzarocks, line 41: We don't want to show the half-second spinner every 10s

simplq/src/components/pages/Admin/index.jsx
```diff
@@ -26,17 +28,27 @@ export default () => {
...
         });
     }
   };
 
+  useEffect(() => {
+    update();
+    return () => clearTimeout(timeoutId);
+  }, [queueId]);
+
+  useEffect(update, [queueId]);
```
daltonfury42, line 50: We don't need two? First one should take care?

thehamzarocks: Yup let me remove the second

daltonfury42:
> If someone approves a PR once, all subsequent commits on the PR don't need approval to merge. This is not how it should be, right? Is there some setting for this?

image

Yes, the dismiss stale requests option should be unchecked. I think you've already done that

daltonfury42:
@thehamzarocks nit: can you declare 10000 as a constant on top?

maaverik:
I see this warning after the new useEffect was added:

.```
/src/components/pages/Admin/index.jsx
  Line 49:6:  React Hook useEffect has a missing dependency: 'update'. Either include it or remove the dependency array  react-hooks/exhaustive-deps

./src/components/pages/QueueStatus.jsx
  Line 61:6:  React Hook useEffect has a missing dependency: 'update'. Either include it or remove the dependency array  react-hooks/exhaustive-deps

Search for the keywords to learn more about each warning.
To ignore, add // eslint-disable-next-line to the line before.
```

It points to the useEffect, i.e.,
 ```
 useEffect(() => {
    update();
    return () => clearTimeout(timeoutId);
  }, [tokenId]);
```

Is it okay to remove the dependencies tokenId and queueId from the useEffects or should I suppress the warning?

thehamzarocks:
I believe removing tokenId from useEffect will make the effect run on every render, which is not what we need.
It's better to suppress warnings or use useCallBack to memoize the function:

https://stackoverflow.com/questions/55840294/how-to-fix-missing-dependency-warning-when-using-useeffect-react-hook

PR #132, opened 2020-07-11
Using the useCallback hook for memoizing functions for useEffect

thehamzarocks:


simplq/src/services/notifications.js
```diff
@@ -0,0 +1,40 @@
+function checkNotificationPromise() {
```
thehamzarocks, line 1: Will use this file as boilerplate for notif permissions later.

For now, submitting PR to prevent warnings in useEffect

maaverik: Could you take care of these warnings before merging please?

thehamzarocks: ? These warnings have been removed through useCallback

simplq/src/components/common/NotificationButton.jsx
```diff
@@ -0,0 +1,71 @@
...
+      if (!('permission' in Notification)) {
+        Notification.permission = permission;
+      }
+
+      dispatch(setNotificationPermission(permission));
+    };
+
+    // Let's check if the browser supports notifications
+    if (!('Notification' in window)) {
+      console.log('This browser does not support notifications.');
```
thehamzarocks, line 34: We could consider doing something more than a console log in this case?

simplq/src/components/pages/Admin/index.jsx
```diff
@@ -28,7 +28,7 @@ export default () => {
 
   const [items, setItems] = useState();
 
-  const update = () => {
+  const update = useCallback(() => {
```
thehamzarocks, line 31: memoizing these functions and passing them as deps to useEffect will remove the warnings and prevent unwanted network calls

simplq/src/services/notifications.js
maaverik: Yeah, all good now. But make sure to replace the materialUI buttons with the ones we made in the common folder later on.

maaverik: And try to make do without Typography, since it's mixing styles with components which we'd prefer not to do.

maaverik: Essentially only use external components like materialUI if it provides some functionality that's difficult to replicate without it. That's why we've kept the old stepper component since it's hard to make and even the textField, since it handles error reporting by itself.

simplq/src/components/common/NotificationButton.jsx
maaverik: We could show this as an error popup. Users won't see the console logs after all

simplq/src/components/pages/Admin/index.jsx
maaverik: What does memoizing the function really mean here? It doesn't return anything so how is it memoized?

thehamzarocks: We need to pass the function ref as deps to useCallback.
If we directly pass a regular function, the reference changes on every re-render, so it would get executed ever time.

On memoization, the function ref changes only when of of the deps specified (queueId here) changes.
So useEffect only runs when queueId (and hence update) changes

maaverik: Ok, I got it

simplq/src/components/common/NotificationButton.jsx
daltonfury42: we can log for now, and come back to this. popup is one option, or we can put some kind of banner or something that this browser/device doesn't support notifications, or w could just stay silent... let's plan it out

simplq/src/components/pages/Admin/index.jsx
daltonfury42: Are you sure this will work? Test

Step 1: Create a queue
Step 2: Add three people, and be in the status page of third user. It will show ahead count as 2.
Step 3: Remove the first user from the queue.
Step 4: In the status page of user 3, press refresh. Confirm that ahead count went down to 1.

simplq/src/components/common/NotificationButton.jsx
daltonfury42: create an isssue

simplq/src/components/common/NotificationButton.jsx
```diff
@@ -0,0 +1,71 @@
...
+import Button from '@material-ui/core/Button';
+import Typography from '@material-ui/core/Typography';
+import { useSelector, useDispatch } from 'react-redux';
+import { setNotificationPermission } from '../../store/appSlice';
+
+const NotificationButton = (props) => {
+  const dispatch = useDispatch();
+  const notificationPermission = useSelector((state) => state.appReducer.notificationPermission);
+
+  const checkNotificationPromise = () => {
```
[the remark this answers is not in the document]
thehamzarocks: This is a way of checking if the browser supports promise-based notification. If there was an exception, it means it's unsupported and we return false.

All this logic is taken straight from 
https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API

simplq/src/components/common/NotificationButton.jsx
```diff
@@ -0,0 +1,71 @@
...
+      console.log('This browser does not support notifications.');
+    } else if (checkNotificationPromise()) {
+      Notification.requestPermission().then((permission) => {
+        handlePermission(permission);
+      });
+    } else {
+      Notification.requestPermission(function (permission) {
+        handlePermission(permission);
+      });
+    }
```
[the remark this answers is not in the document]
thehamzarocks: I agree that this is the old format for functions.
But again, this was taken straight from
https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API.
I can switch to the new arrow functions.

simplq/src/components/pages/Admin/index.jsx
thehamzarocks: Confirmed that this works as expected.

Do you see any reason otherwise?

simplq/src/components/common/NotificationButton.jsx
thehamzarocks: will do.

simplq/src/components/common/NotificationButton.jsx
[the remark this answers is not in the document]
thehamzarocks: Leaving it in the old format for now.

simplq/src/components/pages/QueueStatus.jsx
```diff
@@ -22,26 +23,45 @@ function QueueStatus() {
   const aheadCount = useSelector((state) => state.appReducer.aheadCount);
   const [updateInProgress, setUpdateInProgress] = useState(false);
 
-  const update = () => {
+  const showNotification = useCallback(() => {
+    const notificationImage = '/LogoLight.png';
+    const notificationText = `${queueName}: You've been notified by the queue manager.`;
```
thehamzarocks, line 28: Since we're relying on the queue name to show notifications, this has a cascading effect across our hooks

simplq/src/components/pages/QueueStatus.jsx
```diff
@@ -85,6 +105,7 @@ function QueueStatus() {
       <Header text={queueName} className={styles.header} />
       <JoinerStepper />
       <div>{status}</div>
+      <NotificationButton />
```
thehamzarocks, line 108: Both status and notificationButton are not center-aligned currently.

Whatever we use for centre aligning status gif can be applied here too.

simplq/src/components/pages/QueueStatus.jsx
```diff
@@ -22,26 +23,45 @@ function QueueStatus() {
...
+          }
           timeoutId = setTimeout(update, TIMEOUT);
         })
         .catch((err) => {
           handleApiErrors(err);
           timeoutId = setTimeout(update, TIMEOUT);
         });
     }
-  };
+  }, [tokenId, tokenStatus, dispatch, showNotification]);
```
thehamzarocks, line 59: Any changes to these functions means the memoized update function has to change.
Nice way to avoid subtle bugs.

thehamzarocks:
I've resolved conflicts since the UI refactoring.
Would be nice if this could be merged before any further refactoring is done.

simplq/src/components/pages/Admin/index.jsx
daltonfury42: I was wondering, if the queueId remains the same (which is the case when you click on refresh button), the function will not get called, as we want. can you put a console.log and check?

simplq/src/components/pages/QueueStatus.jsx
daltonfury42: this is fine

simplq/src/components/common/NotificationButton.jsx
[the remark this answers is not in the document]
daltonfury42: lets rename the function to `hasBrowserSupport`

simplq/src/components/common/NotificationButton.jsx
```diff
@@ -0,0 +1,57 @@
...
+      return false;
+    }
+
+    return true;
+  };
+
+  const askNotificationPermission = () => {
+    // function to actually ask the permissions
+    const handlePermission = (permission) => {
+      // Whatever the user answers, we make sure Chrome stores the information
```
daltonfury42, line 23: one thing here

daltonfury42: Can you update the comment, 

```
// Some older versions of chrome doesn't store the permission in the Notification object. In that case we store it manually.
```

simplq/src/store/appSlice.js
```diff
@@ -68,6 +69,10 @@ const appSlice = createSlice({
       // eslint-disable-next-line no-param-reassign
       state.errorText = action.payload;
     },
+    setNotificationPermission: (state, action) => {
+      // eslint-disable-next-line no-param-reassign
+      state.notificationPermission = action.payload;
```
daltonfury42, line 74: lets also store to localstorage?

window.localStorage.setItem('notificationPermission', action.payload);

simplq/src/store/appSlice.js
```diff
@@ -10,6 +10,7 @@ const appSlice = createSlice({
     currentJoinerStep: window.localStorage.getItem('currentJoinerStep'),
     aheadCount: null,
     errorText: '',
+    notificationPermission: Notification.permission,
```
daltonfury42, line 13: see comment below, if it makes sense, then this would be:

```
notificationPermission: window.localStorage.getItem('notificationPermission'),
```

simplq/src/components/common/NotificationButton.jsx
```diff
@@ -0,0 +1,57 @@
...
+
+  if (notificationPermission === 'default') {
+    return (
+      <Button text="Enable Notifications" onClick={() => askNotificationPermission()}>
+        Enable Notifications
+      </Button>
+    );
+  }
+  if (notificationPermission === 'denied') {
+    return <>Enable Notifications in Settings</>;
```
daltonfury42, line 53: even if denied we can let the button be there so that they can enable notifications?

simplq/src/store/appSlice.js
thehamzarocks: Why do we need to do this?

simplq/src/store/appSlice.js
thehamzarocks: window.localStorage.getItem('notificationPermission' ) || Notification.permission

simplq/src/components/common/NotificationButton.jsx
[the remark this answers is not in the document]
thehamzarocks: Ok

simplq/src/components/common/NotificationButton.jsx
thehamzarocks: OK

simplq/src/components/common/NotificationButton.jsx
thehamzarocks: They cannot do that anymore. They'd have to go to browser settings and enable.

simplq/src/components/common/NotificationButton.jsx
thehamzarocks: done

simplq/src/components/common/NotificationButton.jsx
daltonfury42: ah got it.

simplq/src/components/common/NotificationButton.jsx
```diff
@@ -0,0 +1,57 @@
...
```
daltonfury42, line 53: Should it be in <> </> tags, or a <p> tag?

simplq/src/store/appSlice.js
daltonfury42: Only if you see the need to. My thought was, in the old chrome browser scenario that you were telling before, `Notification.permission` will be `undefined`.

So the first time user visits, we will ask for permission and set the user response to `Notification.permission`:

```
      // Some older versions of chrome doesn't store the permission in the Notification object. In that case we store it manually.
      if (!('permission' in Notification)) {
        Notification.permission = permission;
      }
```

I was wondering if the user will again be asked for notification next time he visits, after reopening the browser, if it's not persisted to localstorage.

simplq/src/store/appSlice.js
daltonfury42: see comment below. lets store to local storage only if there is a need to.

simplq/src/components/common/NotificationButton.jsx
maaverik: I think this just defaults to a <p> tag.

daltonfury42: I see.

PR #133, opened 2020-07-11
Refactored code, Made admin and status pages more consistent to design

maaverik:


simplq/src/components/pages/Join/index.jsx
```diff
@@ -39,4 +42,31 @@ export function JoinQueue(props) {
...
+  return (
+    <>
+      <Banner />
+      <InputField
+        placeholder="Enter queue link"
+        onKeyPress={(e) => handleEnterPress(e, () => handleClick(queueLink))}
+        value={queueLink}
+        onChange={(e) => setQueueLink(e.target.value)}
+        className={styles.input}
+        // error={invalidName}
```
thehamzarocks, line 62: Can we remove these comments?

maaverik: I'm keeping it there since we haven't really handled the errors while pasting a link. Ideally, I want this to work with just the name as well the whole link, so we'd need a validator specific to it later on.

PR #146, opened 2020-07-25
Slider for homepage

daltonfury42:


simplq/src/components/common/slider/slider.stories.jsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
```
maaverik, line 1: It should be Slider.stories, right? Capital S like in index.

daltonfury42: Done

maaverik:
Could you add these pictures at the bottom too? That way the page would be complete except for the sidebar.
Screenshot from 2020-07-26 11-45-23

maaverik:
And it would be good if you can reduce the padding here as well.
Screenshot from 2020-07-26 12-26-07

daltonfury42:
All in separate PRs.. The bottom component PR is on it's way. Paddings, text fonts and sizes, I want to take it as a separate one with @akashkbaburajan

PR #148, opened 2020-07-31
Home Page Bottom Icons

daltonfury42:
image


simplq/src/styles/homePage.module.scss
```diff
@@ -50,4 +50,11 @@ div {
   @include text-color();
   @include center-vertically();
   @include center-horizontally();
+  div {
```
thehamzarocks, line 53: These styles will apply to divs globally. Just mentioning. 

Also, I didn’t understand the min width part 🤔

daltonfury42: Without the min-width, when I centre it, it was coming like this, as the length of the text varies: 

image

PR #152, opened 2020-07-31
Admin Refresh and Share Buttons

daltonfury42:
image


simplq/src/components/common/Button/Button.jsx
```diff
@@ -2,14 +2,14 @@ import React from 'react';
...
 
 const StandardButton = (props) => {
-  const { onClick, text } = props;
+  const { onClick } = props;
   return (
     <button
       type="submit"
       onClick={onClick}
-      className={props.className ? props.className : styles['standard-button']}
+      className={props.dark ? styles['standard-button-dark'] : styles['standard-button']}
```
daltonfury42, line 10: rather than allowing class override, which we don't use as of now, I've introduced a dark mode

maaverik: What do you think about creating a dark theme altogether for the site instead of assigning separate light and dark class styles for each component? A theme like what material UI uses. This is kind of what I'm talking about: https://medium.com/@dmitriy.borodiy/easy-color-theming-with-scss-bc38fd5734d1

daltonfury42: ok, so `dark` was maybe a bad name. We have two types of buttons in our design. One is the outlines one, like these:

image

And then there are filled ones:

image

This is different from light/dark theme. I'll rename from `dark` to `outlined`. Then there would be two css classes, `standard-button` and `standard-button-outlined`. Is there a better way to handle this?

maaverik: Yeah, best to rename. We could either use separate classes or create two different button components altogether for both of these, but two components for just the outline seems like overkill. So, let's just use one component with these two classes for now and change later if required.

simplq/src/styles/adminShareBar.module.scss
```diff
@@ -0,0 +1,8 @@
+@import 'utils';
+
+.main-button-content {
+  @include center-horizontally();
+  & > div {
+    margin: 1rem;
+  }
+}
```
maaverik, line 8: Can you put this in the adminPage css file. We don't want to have an entire file for just one class. One file per page is better.

simplq/src/components/common/Button/Button.jsx
daltonfury42: done

simplq/src/styles/adminButtons.module.scss
```diff
@@ -0,0 +1,10 @@
+.main-button-content {
```
maaverik, line 1: Could you rename and move this to the buttons scss script? Since it's only used in the stories script, that makes more sense.

daltonfury42: Done

simplq/src/components/common/Button/Button.jsx
```diff
@@ -2,14 +2,14 @@ import React from 'react';
...
   return (
     <button
       type="submit"
       onClick={onClick}
-      className={props.className ? props.className : styles['standard-button']}
+      className={props.outlined ? styles['standard-button-outlined'] : styles['standard-button']}
     >
-      {text}
+      {props.children}
     </button>
```
maaverik, line 13: This doesn't break any other pages, right?

daltonfury42: It shouldn't. I've tested.

PR #157, opened 2020-08-03
Check if notifications object is available to fix safari compatibility

thehamzarocks:


simplq/src/store/appSlice.js
```diff
@@ -1,5 +1,15 @@
 import { createSlice } from '@reduxjs/toolkit';
 
+// fix for Notification object not supported on iOS safari
+const getNotificationStatus = () => {
+  try {
+    return Notification.permission;
+  } catch (error) {
+    console.log(error);
```
daltonfury42, line 8: can you remove this log line?

PR #160, opened 2020-08-08
Admin JS Code Reorg

daltonfury42:


simplq/src/components/pages/Admin/Item.jsx
```diff
@@ -47,48 +46,48 @@ function Item(props) {
...
         handleApiErrors(err);
       });
     // Notify user of error TODO
   };
   const onDeleteClick = () => {
-    TokenService.remove(tokenId).catch((err) => {
-      handleApiErrors(err);
-    });
-    props.removeItemHandler(tokenId); // TODO Should delete from list only ofter successfull deletion from db
+    props.removeItemHandler(tokenId);
```
maaverik, line 55: Is removing from the tokenService not required anymore?

simplq/src/components/pages/Admin/index.jsx
```diff
@@ -15,52 +14,56 @@ import AddMember from './AddMember';
...
+      .then((data) => {
+        setItems(data.tokens);
+        setQueueName(data.queueName);
+        timeoutId = setTimeout(update, TIMEOUT);
+      })
+      .catch((err) => {
+        handleApiErrors(err);
+        timeoutId = setTimeout(update, TIMEOUT);
+      });
   }, [queueId]);
```
maaverik, line 37: Did you remove the queueId condition because this only fires if queueId gets a valid value?

simplq/src/components/pages/Admin/Item.jsx
daltonfury42: yes, removeItemHandler will call the service, and on success will remove the token from the state as well

simplq/src/components/pages/Admin/index.jsx
daltonfury42: It was some hack I did because queueId was coming from localstorage, and it was coming as null. Now it comes from url. So I can expect it to be present.

Issue #121, opened 2020-07-05
[Design] New UI integrate home page and create queue page


sajmalyousef:
The beginning page currently does not serve any purpose and should have the functionality same as in create queue link.  Simply move the steps section and text box from 2 to 1
1.
![Screenshot from 2020-07-05 11-37-55](https://user-images.githubusercontent.com/13188906/86526668-7d12db80-beb4-11ea-81cc-860f46e7021c.png)

2.
![Screenshot from 2020-07-05 11-42-14](https://user-images.githubusercontent.com/13188906/86526684-a92e5c80-beb4-11ea-8e5b-0634d43d0708.png)

maaverik:
We designed it this way to have a decent looking homepage with the all the information. Putting create functionality there would bloat the page.

sajmalyousef:
@daltonfury42  what do you think? I find it redundant clicking on create a queue just to go to another page to create a queue
Maybe since the steps are already there below in the home page , just introduce text box here ?

akashkbaburajan:
@sajmalyousef a little bit redundancy is fine as long as the idea and message is conveyed.

sajmalyousef:
Steve Krug in his book "Don't make me think" highlights the following attributes of a usable design

Useful: Does it do something people need done?

Learnable: Can people figure out how to use it?

Memorable: Do they have to relearn it each time they use it?

Effective: Does it get the job done?

Efficient: Does it do it with a reasonable amount of time and effort?

Desirable: Do people want it?

and

Delightful: Is using it enjoyable, or even fun?

Often while designing new experiences, we are so convinced that the new design is so much better than the old one in terms of parameters like effectiveness and efficiency, that we forget that users still have to invest in "learning" the new UX. This is one of the reasons why change, no matter how objectively better, is always viewed with skepticism by the customer.  

This is my line of thought. Adding to this another point is:  I've got close to 50 people to use this and our bounce rate is insanely high ---> implying people just never go past the 1st page (with the old UI). With this 1 more navigation we may further affect our chances of getting people to use it. We can maybe discuss further on call.

daltonfury42:
![image](https://user-images.githubusercontent.com/11256376/86528040-1ac0d780-bec2-11ea-8a4a-a53deae48d68.png)

These are my thoughts,

[...]

PR #166, opened 2020-08-08
Refactor Join Page and Status Page JS

daltonfury42:


simplq/src/components/pages/Join/index.jsx
```diff
@@ -14,28 +15,33 @@ import InputField from '../../common/InputField';
 
 export function JoinQueueWithDetails(props) {
   const queueId = props.match.params.queueId;
-  const queueName = useSelector((state) => state.appReducer.queueName);
+  const [queueStatusResponse, setQueueStatusResponse] = useState();
+  useEffect(
+    () => QueueService.getStatus(queueId).then(setQueueStatusResponse).catch(handleApiErrors),
+    [queueId]
```
maaverik, line 21: Does setQueueStatusResponse get called with the status argument here automatically? I didn't realize this syntax worked.

simplq/src/components/pages/Join/index.jsx
```diff
@@ -14,28 +15,33 @@ import InputField from '../../common/InputField';
 
 export function JoinQueueWithDetails(props) {
   const queueId = props.match.params.queueId;
```
maaverik, line 17: I'll split up these components into different files and name them better. I intended to do it later but forgot.

daltonfury42: No lets not. As per https://github.com/SimplQ/simplQ-frontend/issues/121 we would ditch that intermediate page...

And always when you get such ideas during development, raise an issue and move on. It's pretty effective. You get bonus points if you leave a TODO comment in the code as well with a link to the issue.

simplq/src/components/pages/Join/index.jsx
daltonfury42: ha... this is cool... when you have a lambda foo,
```
const foo = (a) => console.log("I'm foo, ", a);
```

and you want to use it in a then, you can do:
```
.then(resp => foo(resp));
```

but then it's the same effect as,
```
.then(foo);
```

Then takes a lambda as calls it with the response.

simplq/src/components/pages/Join/index.jsx
maaverik: Yeah, if that's the path we're going with, then we can ditch this altogether.

simplq/src/components/pages/Join/index.jsx
maaverik: Good to know.

thehamzarocks: This reminds me of method references in Java streams

simplq/src/components/pages/Status/QueueStatus.jsx
```diff
@@ -13,17 +13,16 @@ import QueueDetails from './QueueDetails';
...
-function QueueStatus() {
+function QueueStatus(props) {
+  const tokenId = props.match.params.tokenId;
+  const [tokenStatusResponse, setTokenStatusResponse] = useState();
   const dispatch = useDispatch();
-  const queueName = useSelector((state) => state.appReducer.queueName);
-  const [tokenStatus, setTokenStatus] = useState();
-  const tokenId = useSelector((state) => state.appReducer.tokenId);
-  const aheadCount = useSelector((state) => state.appReducer.aheadCount);
+  dispatch(setJoinerStep(2));
```
thehamzarocks, line 20: I'm wondering if instead of using magic numbers, we should define meaningful constants.

maaverik: Seems like too many imports would be required in a lot of places. If that is the case, what if we just a description of what each number means in the setJoinerStep and setCreatorStep functions atleast as a temporary measure.

daltonfury42: Imports are fine. We should use constants, I feel.

Issue #163, opened 2020-08-08
Explore using redux for handling network calls


daltonfury42:

maaverik:
Could you describe the rationale for adding redux to handle the calls? I thought it was only required when the data coming from the calls is passed around between components a lot which doesn't happen for us, at least for now.

daltonfury42:
Cleaner components. Network calls are not a presentational logic, and hence, I think they should move out. If you look at [Admin page](https://github.com/SimplQ/simplQ-frontend/blob/new-frontend/simplq/src/components/pages/Admin/index.jsx), lines 23-61 can be reduced I feel. I place of all those lines, we can just call actions in the right way.

#170 is also another thing, but I am not sure this is possible, as I have never seen this done before.

daltonfury42:
This works as long as there are just one queue / token being managed. I'm not sure how this will evolve to, if we have more than one token / queue...

daltonfury42:
Actually, before we implement this change, let me talk to someone and validate this...

daltonfury42:
My knowledge on this is a bit outdated, I've seen it being used in the days before `redux-toolkit` like this, using `redux-thunk`:
https://blog.logrocket.com/data-fetching-in-redux-apps-a-100-correct-approach-4d26e21750fc/
https://daveceddia.com/where-fetch-data-redux/

But @maaverik your question still remains, and let me come back to you. Even I'm not 100% convinced

daltonfury42:
@PavanDongare @maaverik I talked to a friend in depth about this, and he also is of the opinion that we don't need to add redux for network calls.

@PavanDongare If there is a strong argument from your side as to why we need this, lets do this. Otherwise lets close the issue.

PR #169, opened 2020-08-09
Style and warning fixes

maaverik:


simplq/src/components/common/Button/Button.stories.jsx
```diff
@@ -10,24 +10,16 @@ export default {
 };
 
 export const CreateQButton = (props) => {
-  return (
-    <StandardButton onClick={props.onClick} outlined>
-      Create Queue
-    </StandardButton>
-  );
+  return <StandardButton onClick={props.onClick}>Create Queue</StandardButton>;
 };
```
maaverik, line 15: We used the wrong className. Outlined means it shouldn't be colored, but we had it the other way around.

simplq/src/components/pages/Status/QueueStatus.jsx
```diff
@@ -50,7 +50,7 @@ function QueueStatus(props) {
         handleApiErrors(err);
         timeoutId = setTimeout(update, TIMEOUT);
       });
-  }, [tokenId, oldTokenStatus]);
+  }, [tokenId, oldTokenStatus, showNotification]);
```
maaverik, line 54: Added this to prevent a warning. Not 100% sure if it's required or not.

simplq/src/styles/buttons.module.scss
```diff
@@ -15,32 +15,22 @@
   &:hover {
     background-color: #fff;
     color: $primary-color-dark;
+    cursor: pointer;
   }
 
   &:focus {
     border-radius: 0.6rem;
   }
 }
```
maaverik, line 25: Renamed to fix the contradiction

simplq/src/components/pages/Status/QueueStatus.jsx
```diff
@@ -50,7 +50,7 @@ function QueueStatus(props) {
...
```
daltonfury42, line 53: Let's leave this warning... if we add `showNotification` here, it will lead to an infinite loop call to TokenService.get().

In the long term, all these non presentational code should move out of our components. For example, notification logic should be managed at a single place (https://github.com/SimplQ/simplQ-frontend/issues/163) and similarly for network calls. (https://github.com/SimplQ/simplQ-frontend/issues/163)

I'm hoping that this will then fix the warning.

simplq/src/components/pages/Status/QueueStatus.jsx
```diff
@@ -98,9 +98,8 @@ function QueueStatus(props) {
         aheadCount={tokenStatusResponse.aheadCount}
       />
       {renderButtons()}
-      <QueueDetails />
+      <QueueDetails queueName={tokenStatusResponse.queueName} />
```
daltonfury42, line 101: let's pass down the entire tokenStatusResponse object... it's of the following structure: 

```
{
    "tokenId": "24895b64-5242-4834-8295-b60f8e9f7037",
    "tokenNumber": 6,
    "tokenStatus": "REMOVED",
    "queueName": "NavaQ",
    "aheadCount": null,
    "notifiable": false
}
```

simplq/src/components/common/Button/Button.stories.jsx
```diff
@@ -38,7 +30,7 @@ export const ShareButton = (props) => {
 
 export const RefreshButton = (props) => {
   return (
-    <StandardButton onClick={props.onClick}>
+    <StandardButton onClick={props.onClick} outlined>
```
daltonfury42, line 33: 🤦 I inverted it, thanks for the fix!

simplq/src/components/pages/Status/QueueDetails.jsx
```diff
@@ -23,7 +26,7 @@ export default () => {
             <p>Creation time:</p>
           </tr>
           <tr>
-            <p>Total number of people in queue:</p>
+            <p>Total number joined so far in queue:</p>
```
daltonfury42, line 29: We have to hit a backend endpoint for getting this information:

```
GET /v1/queue/status?queueId={{queueId}}
```

It currently returns,
```
{
    "queueId": "81abb19b-c2c7-477d-aa53-efe501633743",
    "queueName": "NavaQ",
    "numberOfActiveTokens": 5,
    "totalNumberOfTokens": 6
}
```

So rather than passing the name down, let this component make the network call. We can then have this as an independent component, and use it here, as well as in the Joining page, or even in admin page or somewhere else later on.

daltonfury42: The date is not coming, I'll make the backend change, it should be easy.  https://github.com/SimplQ/simplQ-backend/issues/74

simplq/src/components/pages/Status/QueueStatus.jsx
daltonfury42: Completely ignore the above comment, I've left another comment in the QueueDetails component file.

simplq/src/components/common/Button/Button.stories.jsx
daltonfury42: ha yes, my bad.

simplq/src/components/pages/Status/QueueStatus.jsx
daltonfury42: I've commented, we should remove this for now. Fixing this will make the code ugly.

simplq/src/components/pages/Status/QueueStatus.jsx
maaverik: Okay, I'll remove it

simplq/src/components/pages/Status/QueueDetails.jsx
maaverik: Okay, so right now, we have the tokenId in the QueueStatus script and I need the queueId for this call. I have to use /token/${tokenId} right?

simplq/src/components/pages/Status/QueueStatus.jsx
daltonfury42: I should have done this, but while you are at it, can you leave a comment here so that we don't accidentally put this back?

simplq/src/components/pages/Status/QueueDetails.jsx
daltonfury42: oh I see. In `QueueStatus`, we should be having the `queueId` in `tokenStatusResponse`. It's not currently there. Let me make that change.

daltonfury42: I've made the change. queueId should be present in tokenStatusResponse

simplq/src/components/pages/Status/QueueDetails.jsx
```diff
@@ -1,9 +1,20 @@
...
+import { CircularProgress } from '@material-ui/core';
 import Header from '../../common/Header';
 import styles from '../../../styles/statusPage.module.scss';
+import * as QueueService from '../../../services/queue';
+import { handleApiErrors } from '../../ErrorHandler';
 
 export default (props) => {
   // https://dabblet.com/gist/1506530 --> checkbox hack
+  const [details, setDetails] = useState();
+  QueueService.getStatus(props.queueId).then(setDetails).catch(handleApiErrors);
```
daltonfury42, line 12: This would make the the API call on all re-renders. We should enclose it in a `useEffect` hook, so that it gets called only on changes to `props.queueId`:
```
  useEffect(() => {
    async function fetchData() {
      const response = await QueueService.getStatus(props.queueId).catch(handleApiErrors);
    }
    fetchData();
  }, [props.queueId]);
```

simplq/src/components/pages/Status/QueueDetails.jsx
```diff
@@ -1,9 +1,20 @@
...
+import moment from 'moment';
+import { CircularProgress } from '@material-ui/core';
 import Header from '../../common/Header';
 import styles from '../../../styles/statusPage.module.scss';
+import * as QueueService from '../../../services/queue';
+import { handleApiErrors } from '../../ErrorHandler';
 
 export default (props) => {
   // https://dabblet.com/gist/1506530 --> checkbox hack
+  const [details, setDetails] = useState();
```
daltonfury42, line 11: Lets call this `[queueStatusResponse , setQueueStatusResponse]`

simplq/src/components/pages/Status/QueueDetails.jsx
```diff
@@ -1,9 +1,20 @@
...
+  QueueService.getStatus(props.queueId).then(setDetails).catch(handleApiErrors);
+  if (!details) {
+    return <CircularProgress />;
+  }
+  const timeStamp = moment(details.queueCreationTimestamp);
+  const creationTime = `${timeStamp.format('LT')} ${timeStamp.format('ll')}`;
```
daltonfury42, line 17: Does this take care of timezone difference between server and the user? For example, say server returns UTC time, and client is in India, does this display the converted Indian time?

simplq/src/components/pages/Status/QueueStatus.jsx
```diff
@@ -83,8 +84,15 @@ function QueueStatus(props) {
     return <div />;
   };
 
+  const renderDetails = () => {
+    if (tokenStatusResponse.tokenStatus !== 'REMOVED') {
+      return <QueueDetails queueId={tokenStatusResponse.queueId} />;
+    }
+    return <div />;
+  };
```
daltonfury42, line 92: We can still show the queue details even if it was removed?

simplq/src/components/pages/Status/QueueDetails.jsx
daltonfury42: https://www.robinwieruch.de/react-hooks-fetch-data

maaverik: You're right. Missed this.

simplq/src/components/pages/Status/QueueDetails.jsx
maaverik: Right now, it just converts the time. Let me see if there's a way to fgure out the timezone dynamically.

simplq/src/components/pages/Status/QueueStatus.jsx
maaverik: Should we? My opinion is there's really no point. The queue name is still displayed at the top, but the other details need not be dislayed or updated, right?

simplq/src/components/pages/Status/QueueDetails.jsx
maaverik: It's stored as IST right now in the backend it seems. Can we have it stored as UTC? That way, using solutions like this would make things easy and intuitive 
https://stackoverflow.com/questions/1091372/getting-the-clients-timezone-offset-in-javascript
https://momentjs.com/docs/#/manipulating/utc-offset/

simplq/src/components/pages/Status/QueueStatus.jsx
daltonfury42: Ok. This brings the next point... we need to update QueueStatus in 10s interval... But this is a cross cutting concern that we need to handle at a single place. (Admin Page, Token Status Page and QueueDetails all need to keep refreshing, we don't want to repeat the logic). Pavan would work on this part.

Let's create an issue and move on.

simplq/src/components/pages/Status/QueueDetails.jsx
daltonfury42: Yo. I'll do this weekend. I've raised a BE issue here https://github.com/SimplQ/simplQ-backend/issues/79. Can you also make a FE issue, and let's move on.

simplq/src/components/pages/Status/QueueStatus.jsx
daltonfury42: #170

simplq/src/components/pages/Status/QueueDetails.jsx
daltonfury42: Backend is giving UTC

simplq/src/components/pages/Status/QueueStatus.jsx
maaverik: By saying richer, do you mean having more content on every page?

simplq/src/components/pages/Status/QueueDetails.jsx
maaverik: Really? It shows the right time in IST on the frontend, so I assumed it was storing time in IST. Let me see how moment figures out the timezone.

simplq/src/components/pages/Status/QueueStatus.jsx
daltonfury42: Yes. But this is a product decision. Let's not do it for now.

simplq/src/components/pages/Status/QueueDetails.jsx
daltonfury42: Ah, so backend sends you the UTC time and the timezone is also specified. Moment knows your device's timezone, and might be doing the TZ conversion. That's my guess

maaverik: How does the backend figure out the right timezone if it's not stored? Does it find that out from where the request is sent from? The creationTimeStamp looks like this right now: 2020-08-10T20:23:08.745+0000. I've left a TODO in the code to look at this later. The other stuff is resolved now.

daltonfury42: Backend doesn't know the timezone of the user. Sorry for the confusion. What I had meant was that backend stores the current UTC time. The timezone in the datetime string from backend is "+0000", meaning it's UTC/GMT. For example, backend will send the UI, `2020-08-11T20:22:13.118+0000`

Now moment knows you are in IST timezone, (+0530) and it will add the offset and display the time in IST.

maaverik: Got it. This is probably right.

simplq/src/components/pages/Status/QueueDetails.jsx
```diff
@@ -1,9 +1,31 @@
...
+      setQueueStatusResponse(response);
+    }
+    fetchData();
+  }, [props.queueId]);
+
+  if (!queueStatusResponse) {
+    return <CircularProgress />;
+  }
+
+  // const tzOffset = new Date().getTimezoneOffset();
```
daltonfury42, line 25: remove

maaverik: Will do

PR #173, opened 2020-08-18

jithinlal:


simplq/src/components/common/Button/Button.jsx
```diff
@@ -4,11 +4,7 @@ import styles from '../../../styles/buttons.module.scss';
 const StandardButton = (props) => {
   const { onClick } = props;
   return (
-    <button
-      type="submit"
-      onClick={onClick}
-      className={props.outlined ? styles['standard-button-outlined'] : styles['standard-button']}
-    >
+    <button type="submit" onClick={onClick} className={styles['standard-button']}>
```
daltonfury42, line 7: we neeed to load one of the classes based on `props.outlined`

simplq/src/components/common/Button/Button.jsx
```diff
@@ -4,11 +4,7 @@ import styles from '../../../styles/buttons.module.scss';
...
       {props.children}
```
maaverik, line 8: We still need the outlined buttons for certain pages like for the share and refresh queue buttons, so these shouldn't be removed. The same effect should be added to the 'standard-button-outlined' class as well, with a different transition color. Since this class extends the 'standard-button' class, just adding the different color should be enough.

daltonfury42:
@jithinlal Can you set up local linting... and then can you update the readme with some instructions so that new collaborators would find it helpful?

@PavanDongare had recently solved the local linting issue. You can post here if you need help.

maaverik:
https://blog.echobind.com/integrating-prettier-eslint-airbnb-style-guide-in-vscode-47f07b5d7d6a This will help with the linting setup.

simplq/src/styles/buttons.module.scss
```diff
@@ -10,16 +10,17 @@
...
-  transition: all 0.2s;
+  transition: background 0.8s;
+  background-position: center;
 
   &:hover {
-    background-color: #fff;
-    color: $primary-color-dark;
-    cursor: pointer;
+    background: #776fff radial-gradient(circle, transparent 1%, #776fff 1%) center/15000%;
   }
```
maaverik, line 18: The ripple effect looks good, but I don't really like the color on hover (this is a personal opinion). Maybe something a bit darker to match the primary dark color would be better. @daltonfury42 Any thoughts on the color.

jithinlal:
@daltonfury42 I can see prettier and esline setup already on the code?

simplq/src/components/common/Button/Button.jsx
jithinlal: okay, cool

PR #174, opened 2020-08-18

jithinlal:


jithinlal:
A light shade on hover, no fancy changes, but a good ripple to follow

PR #175, opened 2020-08-18

jithinlal:


jithinlal:
Added a light hue when hovering, and a ripple effect upon click

daltonfury42:
Hi @jithinlal 

The light hue on hovering and the ripple effect is looking nice on the outlined buttons we are having.

In the case of the filled buttons (like the one on homepage), earlier we used to have it turn white+outlined on hover. The current change LGTM. @maaverik How about we merge this, and then we ask @akashkbaburajan and @raima-zachariah for feedback.

daltonfury42:
One more thing, a thin blue outline comes around the last button that was clicked:

image

Can we disable that? (Around the refresh button in this screenshot)

maaverik:
The blue outline is specific to chrome, I've never seen it in firefox. But there must be a way to disable it since I haven't seen this on other sites in chrome. @jithinlal in case you find out how to deal with it, you can add the change later. Other than that, things look good. Great job.

maaverik (approved):

daltonfury42:
I did some googling, and `button:focus { outline: none; }` should do it

daltonfury42:
According to [this](https://medium.com/hackernoon/removing-that-ugly-focus-ring-and-keeping-it-too-6c8727fefcd2) guy, it might affect people using a non-pointer device. He has a solution for it, but in our case, if we do:

```
button:focus { 
  outline: none; 
  // increase the border thickness a bit
}
```
it might take care of that too... @jithinlal what do you think?

jithinlal:
lemme check on this

jithinlal:
closing this pull request to include the border none for chrome browsers.

PR #177, opened 2020-08-19
Ripple effect on button click

jithinlal:


daltonfury42 (approved):
Let's keep updating the same PR with changes, and not close and raise new ones.

@maaverik This change looks good, you can merge

Issue #164, opened 2020-08-08
HandleApiError at one place


daltonfury42:
Hook into axios and catch all backend call failures at one place.

daltonfury42:
https://stackoverflow.com/questions/48990632/how-to-manage-axios-errors-globally-or-from-one-point

daltonfury42:
Resolved by https://github.com/SimplQ/simplQ-frontend/pull/429

PR #189, opened 2020-08-23
Join queue with invalid url

maaverik:


README.md
```diff
@@ -37,4 +37,4 @@ Feel free to fork and improve, and do send a pull request. We will be delighted
 
 There are a ton of features being planned. So if you are considering contributing to this repository, please first discuss the change you wish to make via the issue tracker. Let's work together.
 
-This [readme](/simplq/readme.md) file details the file structure and tools used. Please go through it first.
+This [readme](/simplq/readme.md) file details the file structure and tools used. Please go through it first. The frontend is being revamped heavily right now and the changes are being tested in this branch. We are following this [approach](https://github.com/SimplQ/simplQ-frontend/blob/new-frontend/simplq/readme.md) and implementing this new [design](https://xd.adobe.com/view/8d227f50-0d35-4319-63ad-ff4bad308415-fb13/). You can see the new frontend live [here](https://new.simplq.me/).
```
daltonfury42, line 40: ✌️

simplq/src/components/pages/Join/Form.jsx
```diff
@@ -47,43 +48,51 @@ export function JoinQueueForm(props) {
...
+              onKeyDown={(e) => handleEnterPress(e, handleClick)}
+            />
+          </td>
+        </tr>
+        <tr>
+          <td>
+            {addingInProgress ? (
+              <CircularProgress size={30} style={{ padding: '6px 16px' }} />
+            ) : (
+              <Button onClick={handleClick}>Join Queue</Button>
```
daltonfury42, line 90: Can we use our button here?

simplq/src/components/pages/PageNotFound.jsx
```diff
@@ -1,21 +1,13 @@
...
-          <Button variant="contained" color="primary" onClick={() => props.history.push('/')}>
-            Go home
-          </Button>
-        </Grid>
-      </Grid>
-    </>
+    <div className={styles.main}>
+      <h1>404: Page not found</h1>
+      <Button onClick={() => props.history.push('/')}>Go home</Button>
+    </div>
```
daltonfury42, line 10: I'll raise a low priority jira to style this page...

daltonfury42: #193

daltonfury42: I checked out you branch on local... and it's looking pretty good... Can we also give a button:

"Report a Bug"

that redirects to https://github.com/SimplQ/simplQ-frontend/issues/new ?

simplq/src/components/pages/Join/index.jsx
```diff
@@ -12,13 +12,18 @@ import { Banner } from '../Home/StaticInfos';
...
   const [queueStatusResponse, setQueueStatusResponse] = useState();
+  const [error, setError] = useState(false);
   useEffect(() => {
     async function fetchData() {
-      const response = await QueueService.getStatus(queueId).catch(handleApiErrors);
+      const response = await QueueService.getStatus(queueId).catch((e) => {
+        handleApiErrors(e);
+        setError(true);
+      });
       setQueueStatusResponse(response);
```
daltonfury42, line 27: Few things:

1) This would mean that whatever be the error, the page will redirect to 404. Can we redirect to 404 page only on token not found error from backend?

2) Related to (1), so if we move this logic into handleApiErrors, we can then centrally control places where we want to redirect to a 404 page.. what do you think? (404 page is, semantically thinking a response of handling a certain type of API error)

3) If we move the 404 page rendering logic into handleApiErrors, we can then avoid showing both the snac kbar and the 404 page together, only 404 page would do.


But let's make these changes after @thehamzarocks has done #164 ... deferring this change till then, let's move on... for now this is fine.

daltonfury42 (approved):

simplq/src/components/pages/PageNotFound.jsx
maaverik: Yeah, I'll add this

simplq/src/components/pages/Join/index.jsx
maaverik: For (1) I think it should redirect whatever the error because without the queue info response, the join queue page would make no sense. But like you said, we could think of another message to show instead of 404. Once the error handling is centralized, we can think of passing a heading dynamically to the current PageNotFound component instead of the static 404 heading.

daltonfury42: It should not redirect on all errors.. Errors that are not critical for loading the page need not be redirected. A good example of this is, when you enter a name and click on "Create Queue", the backend will return "The queue already exists, please use a different name". Here the snack bar should do, we don't need the redirect.

But if we are on the queue joining page, and there is no such queue, or same for token status page regarding tokens, then 404 page redirect

maaverik: This is just miscommunication. I was talking about this join queue situation in particular. For other cases, you're right, depends on the context.

daltonfury42: got it now.. then the simple algorithm I had in mind won't work. 

So, on certain API calls, it should go to 404 on any failure... so maybe what you are currently doing is the best way to go? For such calls, just `catch(e -> setError(true))` in the component where the call is made, and redirect to the 404 page. 

---

One thing that came to my mind... here you are setting a state variable "error" to true, and then ...

Instead can we just redirect to the 404 page on line 25?

daltonfury42: In our routes, we have defined the 404 component as a fallback/catch-all route:
```
<Route component={PageNotFound} />
```

so if we redirect to /pagenotfount or something like that... then that would work, if the catchall is working properly..

maaverik: Well the problem with redirecting to the 404 page is that the user can't single click the back button anymore to go back to the old page, since it would immediately redirect if the same error comes up. In this case with the join queue, that's tolerable since going back isn't useful, but we have to think if all cases are like that or not.

maaverik:
I also noticed that the 404 page only takes up half the entire page on the desktop. Does anyone have any ideas how to fill the entire page keeping all the centering intact?

simplq/src/components/pages/Join/index.jsx
maaverik: I'll merge this anyway for now. We can check and merge the other PRs and do testing and think about the error handling parallelly.

daltonfury42: You are right... we can't redirect. 👍

daltonfury42:
@maaverik can you raise an issue for the 404 page not getting vertically centered? It can be a good good first issue

PR #194, opened 2020-08-23
using dispatch hook

PavanDongare:
1) handleApiErrors : doesn't seem to belong to errorhandler component. Had problems using dispatch hook in it
2) ErrorBoundary : seems 2 components declared in same file. 1) errorBoundry 2) ErrorNotification .. gives error on using dispatch.

These 2 needs to resolve before useDispatch here. Rest occurrences covered.

### DIFF
--- a/simplq/src/components/ErrorHandler.jsx
+++ b/simplq/src/components/ErrorHandler.jsx
@@ -8,6 +8,7 @@ import PageNotFound from './pages/PageNotFound';
 import { setErrorNotifOpen } from '../store/appSlice';
 import { store } from '../store'; // TODO: Use Hooks
 
+// ToDo: make functional
 export class ErrorBoundary extends React.Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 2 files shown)


simplq/src/components/ErrorHandler.jsx
```diff
@@ -8,6 +8,7 @@ import PageNotFound from './pages/PageNotFound';
 import { setErrorNotifOpen } from '../store/appSlice';
 import { store } from '../store'; // TODO: Use Hooks
 
+// ToDo: make functional
 export class ErrorBoundary extends React.Component {
   constructor(props) {
```
maaverik, line 13: Do you know if it's possible to get the errorBoundary to work if it's made functional? Granted, it doesn't work right now, but I haven't seen a functional implementation.

maaverik (approved):

simplq/src/components/ErrorHandler.jsx
PavanDongare: I guess making separate component file might help. Will try next.

PavanDongare:
Remaining/ To do : - separating components in errorBoundry file.

simplq/src/components/ErrorHandler.jsx
PavanDongare: @maaverik  after spending an hour trying separating components, came across a statement on react doc site:-
"Only class components can be error boundaries." 😑

maaverik: Ah. Well, that settles it. Now we just have to figure out why it still doesn't work.

PR #203, opened 2020-09-01
Added LoadingIndicator component

maaverik:


maaverik:
Used the PropagateLoader from https://www.davidhu.io/react-spinners/

Screenshot from 2020-09-01 21-33-32

simplq/src/components/pages/Admin/Item.jsx
```diff
@@ -48,7 +48,7 @@ function Item(props) {
     // Notifying in progress
     notificationButton = (
       <IconButton edge="end" color="primary" aria-label="notify">
-        <CircularProgress size={18} />
+        <LoadingIndicator />
```
daltonfury42, line 51: Will these trigger full screen loading?

maaverik: I've made it to show the indicator at the center of whatever component is rendering it. If it is the parent component of the page, it will be full screen, else it shows up at the center of just the part that's loading.

daltonfury42:
This makes sense. 👍

daltonfury42 (approved):
This will be interesting to see in action!

PR #219, opened 2020-09-06
Fix Join Queue Button

daltonfury42:


simplq/src/components/pages/Home/CreateJoinForm.jsx
```diff
@@ -11,7 +11,9 @@ const CreateJoinForm = ({ history }) => {
   const [textFieldValue, setTextFieldValue] = useState('');
   const [invalidMsg, setInvalidMsg] = useState('');
   const [createInProgress, setCreateInProgress] = useState(false);
-  const handleClick = () => {
+  const [joinInProgress, setJoinInProgress] = useState(false);
+
+  const handleCreateClick = () => {
     if (textFieldValue === '') setInvalidMsg('Queue name is required');
     else {
```
maaverik, line 18: We should do the isQueueNameValid check for join and create, right?

daltonfury42: Just checked, the textbox doesn't allow invalid characters. So, for example, if you type "!", it won't go in. So we are good.

maaverik: Oh, got it. The check is there in the handleTextChange function.

maaverik (approved):

PR #225, opened 2020-09-13
PR Automatic Previews

daltonfury42:
Once this is merged, PRs from non-forks will automatically create a preview link.

For example, if you are creating a PR from a branch `abc`, it will create a preview site `abc.simplq.me`

### DIFF
--- a/.github/workflows/preview-teardown.yml
+++ b/.github/workflows/preview-teardown.yml
@@ -1,8 +1,6 @@
 name: 'Amplify PR Closed'
 on:
   pull_request:
-    branches:
-      - master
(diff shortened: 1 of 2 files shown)


maaverik (approved):

maaverik:
This is a great addition. Is there any way to allow this feature for forks as well without compromising security?

daltonfury42:
Running the action on a forked repo has some security risks. Our AWS
credentials are used by this action, which might get leaked. Hence this
limitation.

According to the docs, this action should not be run from forks. If they
do, we will have to revert.

Issue #211, opened 2020-09-05
[HomePage] Features Section Redesign


daltonfury42:
Our features Section looks like this currently: 

![image](https://user-images.githubusercontent.com/11256376/92309776-01bdcb00-efc6-11ea-9e11-a8acf5a91a55.png)

Transform it into: 

![image](https://user-images.githubusercontent.com/11256376/92309789-15693180-efc6-11ea-9356-d7c163ca8f14.png)

Design: https://xd.adobe.com/view/ad1db074-03bf-45b1-537b-98d9d524ec82-db2c/

arkumari2000:
Hello,
I will be happy to do this.
please assign me this issue.

Thank you

maaverik:
Hi @arkumari2000, this item has already been taken care of. Take a look at https://new.simplq.me. The latest code is running here. So closing this item.

PR #239, opened 2020-09-19
Homepage: Why SimplQ Section

daltonfury42:
image


simplq/src/styles/homePage.module.scss
```diff
@@ -28,41 +43,65 @@
...
   }
+  // Position needs to be set for parent so that corona particles can be
+  // positioned absolute to this div
+  position: relative;
+  .corona-1 {
+    height: 7rem;
+    position: absolute;
+    top: 0rem;
+    right: 15vw;
+  }
```
maaverik, line 94: Did you mean rem here?

maaverik:
Screenshot from 2020-09-20 01-12-25

Add a little padding here too.

maaverik:
Screenshot from 2020-09-20 01-15-46

It looks like one of the css changes for the height have messed up the sizes of the admin and join pages.

daltonfury42:
> Screenshot from 2020-09-20 01-12-25
> 
> Add a little padding here too.

It will anyway be completely redesigned in https://github.com/SimplQ/simplQ-frontend/issues/211 



> Screenshot from 2020-09-20 01-15-46
> 
> It looks like one of the css changes for the height have messed up the sizes of the admin and join pages.

I'm not worried, I'm like, we can fix all these when we take up that particular section

simplq/src/styles/homePage.module.scss
daltonfury42: vw, because I wanted it to be relative to the viewport width.

maaverik: Okay

maaverik (approved):

Issue #210, opened 2020-09-05
[HomePage] Create Queue and Join Queue "How To" Section Redesign


daltonfury42:
In the current homepage, on scrolling down, there is a slider titled "Create a Queue at the click of a Button":
![image](https://user-images.githubusercontent.com/11256376/92309569-53655600-efc4-11ea-89ca-96718cf70684.png)

The slider can be removed, and then replaced with three cards like this: 

![image](https://user-images.githubusercontent.com/11256376/92309578-6d069d80-efc4-11ea-9474-26991d5589e2.png)

On wider screens the three cards will be on a single row, but on screen width getting smaller, the cards should flex-wrap down to the next line:
```
flex-wrap: wrap;
justify-content: center;
```

Same for the Join Queue Section:

![image](https://user-images.githubusercontent.com/11256376/92309734-bdcac600-efc5-11ea-8521-60103573572a.png)

Design Link: https://xd.adobe.com/view/ad1db074-03bf-45b1-537b-98d9d524ec82-db2c/

daltonfury42:
Should these cards lift up on hover as is suggested in https://github.com/SimplQ/simplQ-frontend/issues/188 ?

daltonfury42:
Proposal: We should replace with cards as above only bigger screens. On mobile, the slider might be better. 

Rationale: Otherwise, on mobile it would be 8 cards coming on 8 line. It will be long and repetitive scroll. 

We could explore [react-responsive](https://www.npmjs.com/package/react-responsive) for achieving this.

https://medium.com/@kkomaz/react-responsive-its-pretty-cool-d61e5ed56d95

daltonfury42:
Hmm... I want to reduce the number of cards we have on our page. We have 3 rows of cards, total 8 of them.

Can we reduce them... I think we have too much.. Rather than 8 cards, we should just show 4-5 "to the point" ones... We don't have 8 points to convey.

cc: @sajmalyousef @akashkbaburajan @maaverik

daltonfury42:
We can have one row of cards, remaining, we should have maybe a non card svg and two lines of text side by side

daltonfury42:
Another layout idea
![Uploading Screenshot 2020-09-21 at 5.35.12 PM.png…]()

akashkbaburajan:
Yea that's a very valid point. There are just too many cards in the new home page. I'll update the design with the suggestions you have provided

akashkbaburajan:
[...]

PR #245, opened 2020-09-20
Experiment: Slider Tweak

daltonfury42:
@maaverik This slider library that we are using, is a pain to control. The three dots at the bottom was not having proper margin, so I removed it. They also have a autoplay feature, which I enabled, to see what it would be like... What do you think?

If you didn't like autoplay, I can remove that, is it looking like too much motion on the page?

### DIFF
--- a/simplq/src/components/common/slider/Slider.jsx
+++ b/simplq/src/components/common/slider/Slider.jsx
@@ -1,7 +1,10 @@
 import React from 'react';
 import AwesomeSlider from 'react-awesome-slider';
+import withAutoplay from 'react-awesome-slider/dist/autoplay';
 import 'react-awesome-slider/dist/styles.css';
 import styles from '../../../styles/slider.module.scss';
+
+const AutoplaySlider = withAutoplay(AwesomeSlider);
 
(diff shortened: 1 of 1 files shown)


daltonfury42:
Removing this was my main objective:

image

maaverik:
The autoplay doesn't look bad. But the arrows are lost when I use mobile view on the preview so that's a problem. And the autoplay stopped after I enabled mobile view. It doesn't start even after switched back to desktop view. Is it the same on your side?

daltonfury42:
Let me investigate a bit more...

1) The arrows automatically disappear on small screens to save space, it's a feature not a bug.

2) The autoplay stopping on switching modes, I could not reproduce it here. A normal user won't switch like that, so hope that's ok

3) Marking this as WIP. I want to play a bit more
3a) I don't want two autoplaying sliders next to each other... That combined the loaders doesn't look very good.
3b) I want to somehow control it's width.. limited it to some fixed REM.

daltonfury42:
#210 has some examples of layouts, which might be better than a slider... let's not use sliders.

PR #256, opened 2020-09-23
Sentry

maaverik:
Checking if this works


maaverik:
Looks like it works.
Screenshot from 2020-09-24 01-18-13

simplq/src/components/pages/Home/CreateJoinForm.jsx
```diff
@@ -14,6 +14,8 @@ const CreateJoinForm = ({ history }) => {
   const [joinInProgress, setJoinInProgress] = useState(false);
 
   const handleCreateClick = () => {
+    // eslint-disable-next-line no-console
+    console.log(undefined.i);
```
daltonfury42, line 18: let's remove this and check it in.

daltonfury42 (approved):
Interesting thing to test is, I want to intentionally bring the backend down this weekend and then, we should make sure that will give us alert when someone tries to use the site, and gets a 404.

PR #261, opened 2020-09-26
Fix Navbar Opening On Mobile

daltonfury42:
Fix the drawer automatically opening and covering up the homepage on smaller screens


maaverik (approved):

maaverik:
We should add the burger icon too so people are aware that this sidebar exists on phones.

daltonfury42:
@arkumari2000 is working on that. This is a quickfix.

PR #285, opened 2020-10-02
Added Pointer on logo text

Vishv07:


simplq/src/styles/adminPage.module.scss
```diff
@@ -129,5 +129,6 @@
   }
   p {
     font-family: $header-font;
+    cursor: pointer;
```
daltonfury42, line 132: we should add to the img above too.

Vishv07: Yes we can

Vishv07: Okk I added , Have a look

daltonfury42 (approved):

daltonfury42:
Thanks for the PR. And enjoy your tshirt :)

PR #297, opened 2020-10-03
Sidebar For Token Status Page

Gh05d:


simplq/src/styles/sidePanel.module.scss
```diff
@@ -1,10 +1,12 @@
 @import 'utils';
 
 .side-panel {
-  padding-top: 1.5rem;
-  padding-right: 6rem;
+  box-sizing: border-box;
+  padding: 0 6rem 39px 0;
+  display: flex;
+  flex-flow: column;
```
maaverik, line 8: Could you try replacing px with rem? We're using rem consistently everywhere else.

simplq/src/components/pages/Status/NotificationContainer.jsx
```diff
@@ -0,0 +1,66 @@
...
+        <span>Enable notification</span>
+        <InfoOutlinedIcon classes={{ root: 'info-icon' }} />
+      </div>
+    );
+  }
+
+  if (notificationPermission === 'denied') {
+    return <>*If required, please enable notifications in the page settings</>;
+  }
+  return null;
```
maaverik, line 65: Could you try modifying the container so that the toggle is always present instead of this message? On toggling on, the adk function has to be called and if denied, it toggles off.

simplq/src/components/pages/Status/LeaveQueue.jsx
```diff
@@ -1,16 +1,22 @@
...
-    <>
-      <div className={styles['user-action']}>
-        {/* icon */}
-        <div>
-          <h2>Leave Queue</h2>
-          <p>Exit from the queue</p>
-        </div>
+    <li className={styles['user-action']}>
+      <IconButton onClick={() => history.push(`/`)} aria-label="Leave Queue">
+        <CloseIcon />
```
maaverik, line 13: Could you add a props.leaveQueueHandler call here? It is passed in to the sidePanel right now.

  ` <StatusSidePanel leaveQueueHandler={onDeleteClick} queueId={tokenStatusResponse.queueId} />`

You could pass this further into LeaveQueue and call it along with the history.push and I believe that should work.

maaverik (changes requested):

simplq/src/components/pages/Status/LeaveQueue.jsx
```diff
@@ -1,16 +1,22 @@
...
   return (
-    <>
-      <div className={styles['user-action']}>
-        {/* icon */}
-        <div>
-          <h2>Leave Queue</h2>
-          <p>Exit from the queue</p>
-        </div>
+    <li className={styles['user-action']}>
+      <IconButton onClick={() => history.push(`/`)} aria-label="Leave Queue">
```
daltonfury42, line 12: Can we make the entire `<li>` element clickable instead of just the icon? 

We should then change to pointer to hand for `user-action`, and add onHover change background colour to `#f8f9fa`.

simplq/src/components/pages/Status/LeaveQueue.jsx
daltonfury42: I never though of this, but taking the user to home page after `props.leaveQueueHandler` is a really good idea!

simplq/src/styles/sidePanel.module.scss
Gh05d: Done

simplq/src/components/pages/Status/NotificationContainer.jsx
Gh05d: Done. By the way, if the user denies the modal dialoge (sets notifications to denied) then clicking on the switch won't do anything, as `Notification.requestPermission()` does not toggle the modal again (I guess this was done in order to protect the user from spamming). So maybe there should be something like an explanation why this behaves the way it does and tell the user how to activate the notifications.

simplq/src/components/pages/Status/LeaveQueue.jsx
Gh05d: Done

simplq/src/components/pages/Status/LeaveQueue.jsx
Gh05d: I did it and it works, but I run now into these problems with eslint:

  ```
14:5  error  Visible, non-interactive elements with click handlers must have at least one keyboard listener  jsx-a11y/click-events-have-key-events
  14:9  error  Non-interactive elements should not be assigned interactive roles                               jsx-a11y/no-noninteractive-element-to-interactive-role
```

The first one is fixable, the second not. I could turn the whole `li` into a `button` and remove all the styles from it.

Comment (daltonfury42) [reply simplq/src/components/pages/Status/LeaveQueue.jsx:12]: For one let's add a keydown?

Is it possible to replace the li with an anchor or button and style it appropriately? Will this same strategy work for the expansion panel below when we implement that?

The idea is, once you finish this, we should refactor and move this out into a common place and reuse this as a "SidePanel" component here and in the admin page.

File: simplq/src/components/pages/Status/NotificationContainer.jsx (hunk 2 above)
Comment (maaverik) [reply simplq/src/components/pages/Status/NotificationContainer.jsx:65]: I'm not completely sure what you mean. Could you push your latest changes here first? I'll take a look and maybe that'll make things clear.

File: simplq/src/components/pages/Status/LeaveQueue.jsx (hunk 4 above)
Comment (maaverik) [reply simplq/src/components/pages/Status/LeaveQueue.jsx:12]: Making it a button is fine imo, as long as the styles match the design.

File: simplq/src/components/pages/Status/NotificationContainer.jsx (hunk 2 above)
Comment (daltonfury42) [reply simplq/src/components/pages/Status/NotificationContainer.jsx:65]: Right, I forgot about this. Let's leave it like this for now. In a later PR, we should maybe grey it out and then add a tooltip. Can you raise a issue to track this?

Comment (maaverik) [reply simplq/src/components/pages/Status/NotificationContainer.jsx:65]: Also, do try to fix the merge conflicts before pushing.

File: simplq/src/components/pages/Status/LeaveQueue.jsx (hunk 4 above)
Comment (maaverik) [reply simplq/src/components/pages/Status/LeaveQueue.jsx:12]: And the sidepanel common component is already made. After pulling the latest changes, you should see it too @Gh05d.

Comment (Gh05d) [reply simplq/src/components/pages/Status/LeaveQueue.jsx:12]: I went with the _naked_ button solution as it is the easiest way and also the semantically correct one as screen readers will pick it up correctly too.

File: simplq/src/components/pages/Status/NotificationContainer.jsx (hunk 2 above)
Comment (Gh05d) [reply simplq/src/components/pages/Status/NotificationContainer.jsx]: Ok, seems like the conflicts are fixed.

**Comment by maaverik** [comment]
I think something went wrong when you merged the latest changes. I see 42 files changed, and they're mostly from older merged commits. Could you try pulling into the new-fronted branch first and then merging new-frontend to your local version of this branch and pushing again and see if that solves the problem? It's too hard to review right now.

File: simplq/src/components/pages/Status/LeaveQueue.jsx [hunk 5]
```diff
@@ -1,16 +1,23 @@
...
 import styles from '../../../styles/statusPage.module.scss';
 
-export default () => {
+export default (props) => {
+  const history = useHistory();
+  function handleClick() {
+    props.leaveQueueHandler();
+    history.push(`/`);
+  }
+
```
maaverik, line 12: If I leave and then click the back button, I still get the old status page. @daltonfury42, could this be a backend issue?

simplq/src/components/pages/Status/StatusSidePanel.jsx
```diff
@@ -3,11 +3,13 @@ import SidePanel from '../../common/SidePanel';
...
 
 export default (props) => (
   <SidePanel>
-    <LeaveQueue />
+    <LeaveQueue leaveQueueHandler={props.leaveQueueHandler} />
     <ContactAdmin />
     <QueueDetails queueId={props.queueId} />
+    <NotificationContainer />
   </SidePanel>
 );
```
maaverik, line 15: This is just a personal preference, but I like the styles on the admin sidebar better than this one.

Screenshot from 2020-10-04 15-55-55


Screenshot from 2020-10-04 15-55-59

What are your thoughts @daltonfury42  @akashkbaburajan

daltonfury42: I think so too. Let's raise a separate issue to discuss this

simplq/src/components/pages/Status/LeaveQueue.jsx
daltonfury42: It's a front end thing. We can decide on this later.

simplq/src/components/pages/Status/NotificationContainer.jsx
maaverik: The toggle works fine. I prefer letting the user enable notifications later on toggling it on even if they denied once, but I'm not sure how easy it is to implement. We'll decide and try that out later someday.

Gh05d: I would also prefer it that way, but on quick glance over the documentation, I did not find an obvious non hacky way to do it.

Gh05d:
> I think something went wrong when you merged the latest changes. I see 42 files changed, and they're mostly from older merged commits. Could you try pulling into the new-fronted branch first and then merging new-frontend to your local version of this branch and pushing again and see if that solves the problem? It's too hard to review right now.

I tried but it seems like I am doing it wrong. I checked out the new-frontend branch with
```git checkout new-frontend```
did `git pull` to get the most recent version. Then I switched back to my local branch via
```git checkout ticket-291```
and then tried to merge new-frontend into it via
```git merge new-frontend```
Unfortunately, it changed nothing. Is there another way to do this?

**Comment by maaverik** [comment]
@daltonfury42 Any ideas?

**Comment by daltonfury42** [comment]
Can you try:

```
git remote add upstream https://github.com/SimplQ/simplQ-frontend.git
git checkout ticket-291
git pull upstream new-frontend
```

**Comment by daltonfury42** [comment]
I think you mistakenly merged upstream/master branch into your current PR. I don't know a straight forward fix for that, can you selectively undo / restore some files?

File: simplq/src/components/pages/Status/ContactAdmin.jsx [hunk 7]
```diff
@@ -1,16 +1,16 @@
 import React from 'react';
```
daltonfury42, line 1: Just delete all files above this comment in the PR... and manually revert any change that you didn't make, that's the easiest way out.

Gh05d: I just created a new branch and manually pushed my fixes there. Was easier than trying to fix this mess. Sorry for that :-(

daltonfury42:
Closing as duplicate of https://github.com/SimplQ/simplQ-frontend/pull/332

PR #298, opened 2020-10-03
Added Phone number validation

Vishv07:
Now user can't join queue without entering proper number (Length should be 10)


simplq/src/components/pages/Join/Form.jsx
```diff
@@ -77,7 +82,17 @@ export function JoinQueueForm(props) {
                 inputStyle={{
                   width: '100%',
                 }}
-                isValid={() => (invalidContact ? 'Phone number is not valid' : true)}
+                isValid={(inputNumber, country) => {
+                  var arr=[];
+                  arr = inputNumber.split(country.dialCode);
+                  if(arr[1].length!=10){
+                    setInvalidContactLength(true)
```
maaverik, line 89: Make it !== here

maaverik: Also, arr[1].length will cause the app to crash if the user enters one digit and then deletes it. The code would get undefined in arr and would crash at arr[1].length, so that case should be handled.

simplq/src/components/pages/Join/Form.jsx
```diff
@@ -77,7 +82,17 @@ export function JoinQueueForm(props) {
                 inputStyle={{
                   width: '100%',
                 }}
-                isValid={() => (invalidContact ? 'Phone number is not valid' : true)}
+                isValid={(inputNumber, country) => {
+                  var arr=[];
+                  arr = inputNumber.split(country.dialCode);
```
maaverik, line 87: Use const instead of var.

simplq/src/components/pages/Join/Form.jsx
maaverik: The logic used here would work for Indian phone numbers but not other countries. I tried for a Hong Kong number and I can't get a valid number. This isn't an essential thing and can be looked into later if it's too much work.

maaverik (changes requested):

simplq/src/components/pages/Join/Form.jsx
maaverik: The first var arr=[] can be removed safely and const arr = inputNumber... can be used I believe.

daltonfury42 (changes requested):
I'm not very keen on making this change. We should either explore and use an international enterprise ready phone number validation library like [libphonenumber](https://github.com/google/libphonenumber) or leave it unvalidated.

@Vishv07 can you explore https://github.com/google/libphonenumber

simplq/src/components/pages/Join/Form.jsx
Vishv07: Okkk , Than lets keep it as it is

Vishv07:
@daltonfury42  Ya I think you are right , lets keep it as it was

PR #301, opened 2020-10-03
(fix) Button height issue

rasfa98:
### Changes

- [x] Add `min-height: 40px` to button

### DIFF
--- a/simplq/src/styles/buttons.module.scss
+++ b/simplq/src/styles/buttons.module.scss
@@ -12,7 +12,7 @@
   text-align: center;
   transition: background 0.8s;
   background-position: center;
-  min-height: 40px;
+  min-height: 2.5rem;
 
   &:focus {
     outline: none;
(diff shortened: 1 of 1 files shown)


simplq/src/styles/buttons.module.scss
```diff
@@ -12,6 +12,7 @@
   text-align: center;
   transition: background 0.8s;
   background-position: center;
+  min-height: 40px;
```
daltonfury42, line 15: Hi @rasfa98 Thanks for the fix. We prefer to use `rem` throughout. Can you change to `2.5rem`?

daltonfury42 (changes requested):

simplq/src/styles/buttons.module.scss
rasfa98: Will do! 🙂

daltonfury42 (approved):

PR #329, opened 2020-10-04
Smooth scroll to desired section when clicked on navbar links

keithmchd48:
Smooth scroll to desired section when clicked on navbar links (How it Works, Contact Us)

### DIFF
--- a/simplq/src/components/pages/Home/Nav/LeftNav.jsx
+++ b/simplq/src/components/pages/Home/Nav/LeftNav.jsx
@@ -1,18 +1,50 @@
 /* eslint-disable jsx-a11y/anchor-is-valid */
+/* eslint-disable jsx-a11y/no-static-element-interactions */
+
 import React from 'react';
 import styles from '../../../../styles/homePage.module.scss';
+
+function smoothScrollTo(targetElement) {
+  // offset to avoid the sticky header on top to block the "Create a Queue at the click of a button" text
+  const offset = 45;
+  // setting the exact position on the document to scroll to
+  const bodyRect = document.body.getBoundingClientRect().top;
+  const elementRect = targetElement.getBoundingClientRect().top;
+  const elementPosition = elementRect - bodyRect;
+  const offsetPosition = elementPosition - offset;
+  // scroll to the exact position
+  window.scrollTo({
+    top: offsetPosition,
+    behavior: 'smooth',
+  });
+}
+function scrollToHowItWorks() {
+  // get the target div using className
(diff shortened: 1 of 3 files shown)


simplq/src/components/pages/Home/Nav/LeftNav.jsx
```diff
@@ -1,18 +1,49 @@
...
+  // scroll to the exact position
+  window.scrollTo({
+    top: offsetPosition,
+    behavior: 'smooth',
+  });
+}
+function scrollToHowItWorks() {
+  // get the target div using className
+  const element = document.getElementsByClassName('homePage_queue-info__3LUlX')[0];
+  smoothScrollTo(element);
```
maaverik, line 23: Thanks for the PR Keith. I've dome scroll control using React Refs before but not with the actual classname, so I'm a little apprehensive if this would have any undesirable side effects. @daltonfury42 do you have any idea about this?

maaverik: Also, don't worry about the checks for now. They aren't configured correctly yet for external PRs.

maaverik: https://stackoverflow.com/questions/54245395/how-can-i-scroll-to-a-div-onclick-using-react/54245775 This is what I mean by refs btw.

daltonfury42: Using refs might be a pain as there are separate components. Is this possible?

I am also not sure about possible side effects, but if we are going this route, let's assign the divs some unique id and use the id instead of the class name `homePage_queue-info__3LUlX`

keithmchd48: I agree with @daltonfury42. Instead of class names, we should assign unique IDs to represent various sections.

maaverik: I was thinking of creating refs in the parent homepage or layout components and passing to the children. I agree it's a pain.

maaverik: We'll go with the IDs or now. @keithmchd48, can you add the IDs as well?

keithmchd48: Yes I'll do that right away.

keithmchd48:
@daltonfury42 @maaverik I have used ID to target the scroll-to element instead of className in the latest commit.

daltonfury42 (approved):
Thanks for the change. @maaverik you can merge if this looks fine.

Issue #304, opened 2020-10-03
Add Member Expansion Panel Smoothness


daltonfury42:
We have an expansion panel for adding members. Make the open/close transition smooth

![transition](https://user-images.githubusercontent.com/11256376/95001110-94ad4d80-05e4-11eb-98b5-b5733d3c6148.gif)

peakercope:
hey @daltonfury42 
I can help with this if you want

daltonfury42:
Hi @peakercope Please do. Thanks!

daltonfury42:
@peakercope Thanks for fixing this

peakercope:
you're welcome!
let me know if there is anything else where I can help

PR #330, opened 2020-10-04
Add Member Expansion Panel Smoothness

peakercope:
Relates to #304

### DIFF
--- a/simplq/src/components/pages/Admin/AddMember.jsx
+++ b/simplq/src/components/pages/Admin/AddMember.jsx
@@ -10,11 +10,13 @@ export default (props) => {
     setOpen(!open);
   };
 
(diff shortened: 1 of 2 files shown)


simplq/src/components/pages/Admin/AddMember.jsx
```diff
@@ -38,13 +33,15 @@ export default (props) => {
         <ExpandButton />
```
daltonfury42, line 33: Can you also switch the arrow up and down, it would be nice.

peakercope: sure. will do that

daltonfury42 (approved):

PR #333, opened 2020-10-04
Admin Header Styling

Gh05d:
Some comments:

The clickable logo should be indicated as a button to the user, so that it can be focused by tabbing and screen-readers understand it's purpose properly.

The mobile version works fine until 360px, then it kind of breaks. The question is whether even smaller devices should be supported.

My personal opinion is that it is easier to build a website mobile first and styles when the page _grows_. As you can see in my media-query, I just remove padding and margin and it can become hard to understand where everything is coming from when this logic grows.


simplq/src/components/pages/Admin/index.jsx
```diff
@@ -66,7 +68,15 @@ export default (props) => {
 
   const HeaderSection = () => (
     <div className={styles['header-bar']}>
-      <Header className={styles['header']}>{queueName}</Header>
+      <div className={styles['header-title']}>
+        <Header className={styles['header']}>Rooftop Cafê Queue</Header>
```
daltonfury42, line 72: Let's keep both title and subtitle as `{queueName}`

daltonfury42:
@Gh05d We are mostly backend engineers trying to take a stab at frontend. We are not much skilled at CSS, we initially tried designing mobile first, and it looked horrible on desktop. This is us redoing it desktop first, and we are hoping that we can make things to flex-wrap into place, and then tweak it further post October. We are still learning.

We might do a second iteration, post October, and we might take help from our more seasoned web developer friends. 

Post 360px it's not horrible, so I'm ok leaving it like that and revisiting later.

simplq/src/components/pages/Admin/index.jsx
Gh05d: So the App name should not be displayed? I basically tried to implement the design from this sheet.

daltonfury42: We should be displaying the queue name in title. 

Anyone can create a queue. "Rooftop Cafê Queue" was just an example queue name in the design

Gh05d: Ah okay, sorry, I misunderstood that. So is there a prop I could use as a description for the queue?

daltonfury42: It doesn't come from the backend yet. Lets either keep both the title and description as queueName, or we can add a placeholder text for now

Gh05d: I added a description state variable with placeholder text.

Gh05d:
> @Gh05d We are mostly backend engineers trying to take a stab at frontend. We are not much skilled at CSS, we initially tried designing mobile first, and it looked horrible on desktop. This is us redoing it desktop first, and we are hoping that we can make things to flex-wrap into place, and then tweak it further post October. We are still learning.
> 
> We might do a second iteration, post October, and we might take help from our more seasoned web developer friends.
> 
> Post 360px it's not horrible, so I'm ok leaving it like that and revisiting later.

You can open a new ticket and I can try to fix it for smaller devices.

daltonfury42 (approved):

Issue #303, opened 2020-10-03
Copy button expanding on clicking


daltonfury42:
The copy to clipboard is increasing in height when clicked. This is happening due to the height of the new icon on changing.

We need to fix the height of both the icon to be same as other icons

![copy](https://user-images.githubusercontent.com/11256376/95001046-fb7e3700-05e3-11eb-9fa1-8e9197c55169.gif)

nikldev0:
Hi, could I have a go at fixing this?

daltonfury42:
Sure, you are welcome

daltonfury42:
@nikldev0 has fixed the hight, but still the button is expanding in width:

![ezgif-7-96b5bce3618c](https://user-images.githubusercontent.com/11256376/95117643-88e99480-0766-11eb-9824-b69ccdabea34.gif)

If possible, can we add a fixed with to just this button?

sofistoyanova:
can i take it?

maaverik:
Sure. Go ahead.

mubashiralisiddiqui:
Pull request created for this issue please review it
https://github.com/SimplQ/simplQ-frontend/pull/375

PR #339, opened 2020-10-04
Fixed Button Expansion Issue

nikldev0:
For "Copy button expanding on clicking #303"

### DIFF
--- a/simplq/src/styles/buttons.module.scss
+++ b/simplq/src/styles/buttons.module.scss
@@ -4,7 +4,7 @@
   background-color: $primary-color-dark;
   color: #fff;
   display: inline-block;
-  padding: 0.2em 0.5em 0.3em 0.5em;
+  padding: 0.2em 0.5em 0.3em 0.2em;
   font-weight: bold;
   font-size: large;
   border: 2px solid $primary-color-dark;
@@ -50,11 +50,14 @@
   display: flex;
   align-items: center;
   max-width: 14rem;
-  min-height: 1rem;
+  max-height: 1.1rem;
   & > * {
     margin: 0.2rem;
   }
 }
+
+
+
 
 .fork-on-github-button {
   text-decoration: none;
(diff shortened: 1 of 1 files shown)


daltonfury42:
@nikldev0 I think we should rather fix the size of the Tick mark icon

nikldev0:
> 
> 
> @nikldev0 I think we should rather fix the size of the Tick mark icon

Sure @daltonfury42 let me change that and revert back to the min-height property

nikldev0:
@daltonfury42 done thanks, pls check

simplq/src/styles/buttons.module.scss
```diff
@@ -4,7 +4,7 @@
   background-color: $primary-color-dark;
   color: #fff;
   display: inline-block;
-  padding: 0.2em 0.5em 0.3em 0.5em;
+  padding: 0.2em 0.5em 0.3em 0.2em;
```
daltonfury42, line 7: Why did we reduce this padding?

nikldev0: I thought it looked a little odd with the additional space to the left of the tick icon so I reduced it, but will revert it if you would like?

daltonfury42: I see, let's see how it looks.

simplq/src/styles/buttons.module.scss
```diff
@@ -29,6 +29,10 @@
   }
 }
 
+.standard-button img {
+  height: 24px;
```
daltonfury42, line 33: Can you make this `1.5rem`? We prefer rem over px

simplq/src/styles/buttons.module.scss
```diff
@@ -56,6 +60,9 @@
   }
 }
 
+
+
+
```
daltonfury42, line 65: remove

daltonfury42 (approved):

daltonfury42:
Thanks for the PR!

PR #345, opened 2020-10-04
Fix Admin Sidebar Collapse Issue 

daltonfury42:


simplq/src/components/common/Footer.jsx
```diff
@@ -59,9 +59,7 @@ export default () => (
           <div className={styles['list-container']}>
             <ul>
               <li>
-                <a target="_blank" href="https://iimb.qualtrics.com/jfe/form/SV_aY8FY91ztRX9NvD">
-                  Contact Us
-                </a>
+                <a href="https://iimb.qualtrics.com/jfe/form/SV_aY8FY91ztRX9NvD">Contact Us</a>
```
daltonfury42, line 62: removing unwanted target="_blank"  here and below

we don't want to open in a new link

maaverik (approved):

Issue #327, opened 2020-10-04
Standardise and Give a pagewide margin


akashkbaburajan:
Compared to design, website is cramped up in the middle. All sections need to be expanded so that blank space is not too much.
Top part is design and bottom is current website
![Adobe XD 04-10-2020 13_18_49 (2)](https://user-images.githubusercontent.com/16539463/95011573-441f0a00-064f-11eb-99bd-bc51946ae52a.png)

daltonfury42:
Yes lets increase spacing, especially margins. Related discussion https://github.com/SimplQ/simplQ-frontend/issues/323

daltonfury42:
Currently we have individual margin for each section. We should have one left and right margin for the whole page defined on the body or so. Then we should give negative margins for for components that go out (like the divs with a background colour in out design.

karthiksista:
I would like to work on this.
Can I pick this one?

daltonfury42:
@karthiksista This task might overlap with https://github.com/SimplQ/simplQ-frontend/issues/323 which someone else is already working on.

Would you like to take up the similar task for admin page instead? https://github.com/SimplQ/simplQ-frontend/issues/325

karthiksista:
@daltonfury42 Ok sure. Will Do.
Thanks

daltonfury42:
Raised PR for home page https://github.com/SimplQ/simplQ-frontend/pull/378

Issue #325, opened 2020-10-04
Admin dashboard fix margin and spacing


sajmalyousef:
The current admin dashboard is spread far apart compared to original design. It looks a bit zoomed (1st image) and not as good as the  design ( 2nd image)
![image](https://user-images.githubusercontent.com/13188906/95011118-0f5d8380-064c-11eb-90e7-794f5ada458f.png)

![image](https://user-images.githubusercontent.com/13188906/95011148-492e8a00-064c-11eb-9437-c09a744675db.png)

Can it be centered maybe?
https://github.com/SimplQ was looking at this page , wherein things are centered a bit

akashkbaburajan:
Yes, 2 other points I wanted raise are:
1. There are a few spacing issues compare to design for all those queue setting tiles on right side - spacing between "+" to "Add User Manually" to text under it.
2. Add user manually expansion should happen on clicking anywhere on the tile and not just on the "v"

akashkbaburajan:
3. Refresh queue makes more sense to me than refresh status.
4. colour of refresh queue and copy to clipboard buttons should change
5. add user manually makes more sense than
![Uploading Adobe XD 04-10-2020 14_43_54 (2).png…]()
 add member

akashkbaburajan:
3. Refresh queue makes more sense to me than refresh status.
4. colour of refresh queue and copy to clipboard buttons should change
5. add user manually makes more sense than add member
![Adobe XD 04-10-2020 14_43_54 (2)](https://user-images.githubusercontent.com/16539463/95011742-8eed5180-0650-11eb-87bb-18086a86b082.png)

daltonfury42:
We should standardise and use the same margin that we use in home page (See https://github.com/SimplQ/simplQ-frontend/issues/327)

For the buttons, I feel the white background looks better, and also that's easier from implementation standpoint (Changing this would add extra complexity). So can we keep them as they are?

We should fix all the other points raised

akashkbaburajan:
those colour differences were made to make the important button more catchy. I guess its fine to leave them white for now

karthiksista:
I will be working on this.

karthiksista:
Created a [PR](https://github.com/SimplQ/simplQ-frontend/pull/347) for this issue.

daltonfury42:
Done

PR #347, opened 2020-10-05
Fix admin dashboard

karthiksista:
1. @daltonfury42 As per your comment on this issue 

> We should standardise and use the same margin that we use in home page (See #327)

As #327 is still in progress, i have set a `margin` of `10rem `as of now in the admin page.

2. > There are a few spacing issues compare to design for all those queue setting tiles on right side - spacing between "+" to "Add User Manually" to text under it.

Fixed

3. > Add user manually expansion should happen on clicking anywhere on the tile and not just on the "v"

Fixed

4. > Refresh queue makes more sense to me than refresh status, add user manually makes more sense than add member 

Test changes fixed.

5. > those colour differences were made to make the important button more catchy. I guess its fine to leave them white for now.

I have left the background  color as white 

admin-page

### DIFF
--- a/simplq/src/styles/adminPage.module.scss
+++ b/simplq/src/styles/adminPage.module.scss
@@ -247,7 +247,9 @@
   display: flex;
   flex-wrap: wrap;
   align-items: center;
-
+  @media screen and (max-width: 400px) {
+    padding: 0 8.3rem;
+  }
   & > img {
     height: 3rem;
     width: 3rem;
@@ -277,10 +279,21 @@
     .main-button-group {
       margin: 0;
       padding-right: 0;
+      display: flex;
+      flex-direction: row;
+    }
+  }
+
+  .header-title {
+    .sub-header {
+      min-width: 150%;
     }
   }
   .main-body {
     margin: 1rem 2rem;
+  }
+  .navbar {
+    padding: 0 8.3rem;
   }
   .admin-action {
     .collapsable {
--- a/simplq/src/styles/adminPage.module.scss
+++ b/simplq/src/styles/adminPage.module.scss
@@ -1,11 +1,13 @@
 @import 'utils';
 
 .admin-action {
+  @include naked-button();
   font-size: 0.8rem;
   display: flex;
   align-items: center;
   padding: 0;
   min-width: 15rem;
+  text-align: left;
   color: $primary-color-dark;
 
   & > * {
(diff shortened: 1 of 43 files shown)


karthiksista:
admin-page

simplq/src/styles/adminPage.module.scss
```diff
@@ -29,8 +33,9 @@
       flex-direction: row;
       align-items: center;
       .expand-icon {
-        margin-left: 2rem;
+        margin-left: 7rem;
```
daltonfury42, line 36: Why did we add this margin?

daltonfury42 (approved):

daltonfury42:
Can you use media queries and set margin to 1 or 2 rems on smaller screens?

simplq/src/styles/adminPage.module.scss
karthiksista: As per the suggested layout in the [issue](https://github.com/SimplQ/simplQ-frontend/issues/325), the chevron is placed to the right of the div.
As the Previous layout was a little bit zoomed in the margin of 2rem's was sufficient to place the chevron/ expand icon to the right.

If we did not add the extra margin, it would look like the image below
chev-without-margin

karthiksista:
> Can you use media queries and set margin to 1 or 2 rems on smaller screens?

@daltonfury42 Yes, I can but not sure where you want that to be implemented particularly!

karthiksista:
I have added media querie, pls verify. Attaching a screenshot of a mobile view.
mob-view

daltonfury42:
Lets reduce all side margins down: title section and footer too, on mobile view. Thnaks!

karthiksista:
> Lets reduce all side margins down: title section and footer too, on mobile view. Thnaks!

Sure Will do.

karthiksista:
Pushed the requested changes, Pls verify.
Thank You.

simplq/src/styles/adminPage.module.scss
daltonfury42: I don't think this is the right solution. This has been solved independently in another PR. Can you pull the latest changes?

simplq/src/styles/adminPage.module.scss
```diff
@@ -223,21 +233,23 @@
...
+  margin: 3rem 10rem;
 }
 
 .navbar {
   width: 100%;
   height: 4rem;
   background-color: #fff;
   border-bottom: 0.2rem solid #f1f1f1;
-  padding: 0 3rem;
+  padding: 0 11rem;
```
daltonfury42, line 244: I am having trouble making sense of these margins, I'll have to pull the code locally and see, before I can approve these changes.

Can you pull our latest code and fix conflicts?

daltonfury42 (changes requested):
.

karthiksista:
> see comments above

I have taken the latest pull, and resolved merge conflicts.

maaverik:
@daltonfury42, I'm not super convinced if this approach is correct for this issue since this is just for the admin page and not the whole site. What do you think about waiting for a while, so we can think through whether this is the right approach we want to take?

daltonfury42:
@karthiksista Please see this PR: https://github.com/SimplQ/simplQ-frontend/pull/378

I don't want those hacky paddings and margins.

Issue #296, opened 2020-10-03
Status page: Logo should redirect to the home page


Vishv07:
![logotohomepage](https://user-images.githubusercontent.com/37216460/94991204-22fee080-059f-11eb-864b-d5f8578db3d2.JPG)

There's no action after clicking on logo , logo should redirect to the homepage

maaverik:
Hey @Vishv07, this status page is undergoing a redesign and this issue will be taken care of during this redesign. Take a look at #282 to understand more. I'll modify this issue to add this redirection in the new design in a little while. If you want, you can try working on it.

maaverik:
The header on the status page should have the logo and simplq on it as shown in the image below. These should redirect to the homepage on click.

![Screenshot from 2020-10-03 19-18-42](https://user-images.githubusercontent.com/12583528/94993288-84c64700-05ad-11eb-90cb-25891d8e8a93.png)

This logo has already been used in the admin and home page navbars and so could either be converted into a component of it's own and used here or if that's a little too difficult, you can make a copy of the code for now and we can refactor later.

Vishv07:
@maaverik So are you trying to say we should put this whole header on the status page.

maaverik:
The header is already there. The logo and image on the left aren't.

Vishv07:
Okk I'll make it done , will open PR soon

maaverik:
I've opened #328 to make the logo reusable. Maybe hold off till that's merged. Then you can reuse it here and make required styling changes.

Vishv07:
Cool , Will do that

iampankajgupta:
can i work on this issue

Vishv07:
@iampankajgupta i am already working on It , Probably will open PR today 👍

maaverik:
Thanks for the PR @Vishv07. Great job.

Issue #331, opened 2020-10-04
{Design] Join queue page on mobile


maaverik:
![Screenshot from 2020-10-04 16-27-18](https://user-images.githubusercontent.com/12583528/95013559-ab908600-065e-11eb-87f2-820fd73a725e.png)

The join page isn't really suitable for mobile view right now. If we just use a flex, the form would either show up before the header or after the details (which would make the user have to scroll down to join) @akashkbaburajan could you think of a suitable mobile view for this?

daltonfury42:
We should use the same general layout of the status page if possible. It will give the user a consistent feel and also we will be able to reuse a lot of code

https://xd.adobe.com/view/ad1db074-03bf-45b1-537b-98d9d524ec82-db2c/screen/e87838cd-febc-4f37-a68a-9daa0d6b207b/specs/

akashkbaburajan:
yes, I have redesigned the page and updated the design link. How does this look
![Adobe XD 04-10-2020 19_42_00](https://user-images.githubusercontent.com/16539463/95018016-6bd79780-067a-11eb-8b13-1c62d0764dea.png)

daltonfury42:
I'm liking this

maaverik:
Let's use this design and make a new issue to create this page? The header is being worked on anyway right now in #348. @akashkbaburajan can you provide the XD link for this page?

daltonfury42:
I'd suggest splitting to three tasks:

Task 1: Refactor header to common folder and then just use it in the join page.
Task 2: Style left Pane
Task 3: Style right pane

Where task 1 is blocked on #348 and task 2 and 3 are "You can do this". But. I know you don't like splitting to subtasks :)

maaverik:
Task 2 is pretty small, nothing much to be done there. Task 3 would take more work and maybe an extra API call. Could you make an issue on task 3 detailing the API which would need to be used? I guess the same one as in QueueDetails. As for splitting, I don't mind either way. Whoever's doing task 3 can do task 2 easily. I'll leave that up to you :).

daltonfury42:
Can I do it in the weekend?

Another thought: We are using a 3rd party phone number input and material UI's text box. Changing their styling might be pain. Can we stick to the input boxes that we have, for now?

maaverik:
[...]

PR #348, opened 2020-10-05
Redesigned the header for join-queue page

Vishv07:
used common logo component for header  
as we have discussed on  #296


maaverik:
#296 is an issue intended for the status page (src/components/pages/Status/QueueStatus.jsx), but the changes here are for the join page. This is still fine since the join page header is also similar to that of the status page (#331), so happy accident :). 

Two things to note. One is that the image color is a problem. It doesn't match with the design. We might need to create another logo component with this other color for the status and join pages. This is something you could do. You can find the asset here and modify the ClickableLogo component to make it give different colored images according to a prop (imageColor = dark/light).

The second thing is that the logo should be on the same line as the header on mobile view. It should not wrap like this.
Screenshot from 2020-10-05 21-20-59

maaverik:
Once you make the logo variation, you can add it to the status page as well.

simplq/.eslintrc.json
```diff
@@ -7,7 +7,9 @@
     "node": true
   },
   "rules": {
-    "prettier/prettier": ["error"],
+    "prettier/prettier": ["error", {
+     "endOfLine":"auto"
+   }],
```
maaverik, line 12: What does this do?

Vishv07:
@maaverik Okay , I'll have look soon , Seems like these are some minor changes

simplq/.eslintrc.json
Vishv07: This is for VScode , It was giving some error in Editor so I just turned it off

simplq/src/components/pages/Join/index.jsx
```diff
@@ -45,8 +46,17 @@ export default function JoinQueueWithDetails(props) {
 
   return (
     <div>
-      <SimplQHeader />
-      <Header className={styles.header}>{queueStatusResponse.queueName}</Header>
+      <div className={styles['header-bar']}>
+        <div className={styles['simpleq-logo']}>
+          <Logo history={props.history} />
```
maaverik, line 51: Rename to simplq-logo

simplq/src/styles/joinPage.module.scss
```diff
@@ -13,3 +17,102 @@
...
+.main-body {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  margin-bottom: 3rem;
+}
+
+.navbar {
+  width: 100%;
+  height: 4rem;
```
maaverik, line 76: This class is not required.

simplq/src/styles/joinPage.module.scss
```diff
@@ -13,3 +17,102 @@
...
+    font-size: 2.7rem;
+    margin-bottom: 0;
+  }
+
+  .admin-button {
+    padding: 0 0.1rem;
+  }
+
+  .main-button-group {
+    @include center-horizontally();
```
maaverik, line 53: This is also not used anywhere it seems. Remove the unused classes.

Vishv07: I just copied those css from main header component,  forget to remove it , will make clear in next PR. Thanks

simplq/.eslintrc.json
maaverik: Don't change the config directly. In case you have an error you can't figure out, add an eslint comment to ignore the error for just that line or file. We can look at that later then.

Vishv07:
@maaverik have a look

simplq/.gitignore
```diff
@@ -18,6 +18,9 @@
 .env.test.local
 .env.production.local
 
+
+#json
+.eslintrc.json
 npm-debug.log*
```
maaverik, line 24: We can't ignore eslintrc. A change might be required later that should be followed by everyone. So remove it from gitignore, and remove the endOfLine change and make VSCode ignore the error you're getting just within that particular file.

maaverik: Screenshot from 2020-10-06 13-15-22

Vishv07: In My VScode its showing error at everyline. Thats why i did that . but okay I'll keep that as it was

Vishv07: Check now

simplq/src/styles/joinPage.module.scss
```diff
@@ -13,3 +17,39 @@
...
+.header-bar {
+  @include background-color();
+  padding: 0.5rem;
+  // padding-top: 4rem;
+  
+  display: flex;
+  flex-wrap: wrap;
+  .simplq-logo{
+    margin-top: 35px;
+  }
```
maaverik, line 41: Can you replace the px with rem? We prefer using rem as the sizing unit everywhere.

maaverik: And also add some more padding to the left in order to match the design.

Screenshot from 2020-10-06 13-25-48

maaverik: These are minor thing, but the heading is not correctly centred it seems.
Screenshot from 2020-10-06 13-27-18

simplq/src/styles/joinPage.module.scss
```diff
@@ -13,3 +17,39 @@
...
+  .header {
+    @include standard-header();
+    font-size: 2.7rem;
+    margin-bottom: 0;
+  }
+}
+
+@media screen and (max-width: 1000px) {
+  .header-bar {
```
maaverik, line 50: rem here too please.

Vishv07: Okkk I'll change it soon , I'll keep that in mind to use rem. 
And that name and description is looks perfect (centered) in my machine what could be the issue. The screenshot that you have posted has very different than mine .

I'll try to figure it out

Vishv07: I think px would be fine for media queries , for general rem would be fine

maaverik: Yeah I suppose that's fine.

maaverik:
Good work @Vishv07. The centering looks fine now. The only problem I see is that the logo overlaps with the header on mobile view. Ideally it should stay on the same line as long as there is room for it and then flex. It should never overlap. I said it should stay on the same line in mobile view before. That was my bad. I wanted to say it should stay on the same line if there is room for it, else there's no other option than to wrap. Sorry for the confusion.

Could you please add the wrap back, preferentially with the logo on the aligned left even if it shows up above the header, unlike the first image I posted where the logo was centered on wrap?  You'll have to remove the left padding in the moblie specific view for this I think. Try that out. 

Screenshot from 2020-10-06 22-51-47

Vishv07:
I felt that too , But that's fine. how about this ? is that fine

image

maaverik:
Yes, this looks fine. Do you have any more suggestions @daltonfury42?

simplq/src/components/common/ClickableLogo.jsx
```diff
@@ -10,10 +10,20 @@ export default (props) => {
...
+    );
+  }
+  if (props.type === 'Dark') {
+    return (
+      <div className={styles['logo']} onClick={onClick}>
+        <img src="/Simple-Q.png" alt="Home" />
+        <p>SimplQ</p>
+      </div>
+    );
+  }
```
daltonfury42, line 28: Can we avoid this repetition?
```
    return (
      <div className={styles['logo']} onClick={onClick}>
        <img 
          src={props.type.toLocwerCase === 'light' ? "/LogoLight.png" : "/Simple-Q.png"} 
          alt="Home" 
         />
        <p>SimplQ</p>
      </div>
    );
```

simplq/src/components/pages/Admin/index.jsx
```diff
@@ -95,7 +95,7 @@ export default (props) => {
   const Navbar = () => (
     <div>
       <nav className={styles['navbar']}>
-        <Logo history={props.history} />
+        <Logo history={props.history} type="Light" />
```
daltonfury42, line 98: lets be consistent with case, change to `light` here and other places.

daltonfury42 (approved):
Thanks for this!! Some minor changes

maaverik:
This latest change is not in yet right, @Vishv07?

Vishv07:
@maaverik nope not yet , but here it is 
@daltonfury42 Okay i did some changes

maaverik (approved):

daltonfury42:
@Vishv07 Good one! We are coming close to perfection one by one.  @maaverik

PR #356, opened 2020-10-05
Implement the queue details collapsible dropdown in status page

### LINKED ISSUE #343: Implement the queue details collapsible dropdown in status page
The queue details button should open a dropdown with the details of the queue as given here.

Screenshot from 2020-10-04 22-16-29

Take a look at the commented part of src/components/pages/Status/QueueDetails.jsx to get an idea of the required data that comes from the backend. This same commented table has to be implemented to fit the above given design. The src/components/pages/Admin/AddMember.jsx file is a good reference to create a collapsible element.

giubatt:
- Added SidePanelButton common component
- Refactored sidebar buttons from admin and queue to use SidePanelButton
- Added queue details info to side panel button
- Removed some styles that became unused after refactor

Closes #343

### DIFF
--- a/simplq/src/components/common/SidePanelButton.jsx
+++ b/simplq/src/components/common/SidePanelButton.jsx
@@ -1,0 +1,63 @@
+import React, { useState } from 'react';
+import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+import { Collapse } from '@material-ui/core';
+import PropagateLoader from 'react-spinners/PropagateLoader';
+import styles from '../../styles/sidePanelButton.module.scss';
+
+const ExpandButton = ({ isOpen, onToggle }) => (
+  <div
+    role="button"
+    tabIndex={0}
+    className={styles['expand-button']}
+    onClick={onToggle}
+    onKeyDown={onToggle}
+  >
+    <ExpandMoreIcon
+      className={
+        isOpen
+          ? `${styles['expand-icon']} ${styles['expand-icon-open']}`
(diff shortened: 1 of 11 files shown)


maaverik:
This looks fantastic @giubatt. Thanks a lot. 

Should we default to keeping the collapsible buttons collapsed @daltonfury42? 

Also, it might be good to have the entire button made clickable instead of just the arrows for collapsible buttons. But I don't know how easy this would be to implement because the add member button has a form component inside. The users would still need to be able to use the form without problems even if the button collapses on click. This can be tried out later I guess.

maaverik (approved):

giubatt:
> This looks fantastic @giubatt. Thanks a lot.
> 
> Should we default to keeping the collapsible buttons collapsed @daltonfury42?

Oh I forgot it set to open by default, was using that while developing, meant to have it closed by default, I'll change it.

> Also, it might be good to have the entire button made clickable instead of just the arrows for collapsible buttons. But I don't know how easy this would be to implement because the add member button has a form component inside. The users would still need to be able to use the form without problems even if the button collapses on click. This can be tried out later I guess.

I can do that, I did that initially but I was unsure how good/bad it would be to use as if you click outside the inputs it would close, but the inputs themselves work fine. I changed it to how it is now because I noticed the current Add Member only expands when clicking the button.

daltonfury42:
This is nice refactoring. Thanks. Yes, let's see later about how to do that. But we should

daltonfury42 (approved):

PR #357, opened 2020-10-06
Added click handler to place call from call icon

sethbrotherton:
Added a click handler to IconButton using native method to place phone call


maaverik:
This looks like it works fine. Thank you @sethbrotherton. Does this look complete @daltonfury42?

daltonfury42 (approved):
Yes

PR #360, opened 2020-10-06
Fix z-indexes for nav and overlay

devashishpuri:


daltonfury42:
Sorry this is a duplicate of #361

maaverik:
This looks like it works. It's doesn't solve the core problem I guess, but I can't think of any side effects right now. @daltonfury42, can you check if you notice any problems with this change?

maaverik:
Looks like this works now. @daltonfury42 could you also check?

daltonfury42:
Respecting the fact that someone else was already working on this, I had closed this PR as a duplicate. (#361)

maaverik:
#361 solved the core problem while this PR was more of a workaround, so we'll go with that one. Thanks for the effort @devashishpuri.

PR #361, opened 2020-10-06
Incorrect z-index fixed

icy-meteor:


daltonfury42 (approved):
I think this makes sense. @maaverik can you take a look?

daltonfury42 (changes requested):
The text becomes unclickable. Can you take a look?

icy-meteor:
Sure, taking a look.

icy-meteor:
Looks like the way navbar elements are laid down is causing some overlaps in margin. I am refactoring the code in a better and cleaner way @daltonfury42

icy-meteor:
Fixed it! Can you take a look now @daltonfury42  :)

maaverik:
Looks like this works now. @daltonfury42 could you also check?

simplq/src/components/pages/Home/Nav/LeftNav.jsx
```diff
@@ -12,24 +12,26 @@ const scrollToHowItWorks = () => {
...
+          <a>My Queues</a>
+        </li>
+        <li>
+          <a tabIndex={-1} href="https://iimb.qualtrics.com/jfe/form/SV_aY8FY91ztRX9NvD">
+            Contact Us
+          </a>
+        </li>
+        <li>
+          <a>Sign In/Sign Up</a>
+        </li>
```
maaverik, line 32: Is there any change here? Or was it just removing the left in the scss file that fixed this?

icy-meteor: Also, wrapped the entire <ul> under a <div> tag in LeftNav.jsx component to get it working :)

daltonfury42:
I've generated a preview here: https://click.dpmhbfuiok7f8.amplifyapp.com/

daltonfury42 (approved):
I can confirm that the requirements are met.

simplq/src/components/pages/Home/Nav/LeftNav.jsx
daltonfury42: +1 How did this fix the isssue?

icy-meteor: There are multiple ways to fix this. Ideally, it is something to do with the way CSS positioning has been used. Wrapping in a div, would ensure that anything around the div is consuming its entire width and height, giving space for other div's as well. In this case, the `<ul>` was not wrapped in a div and the browser couldn't calculate exact width consumed by it, hence the left property pulled it on top of the logo text :)
I know it sounds confusing, but hope it helps.

maaverik: Okay, I didn't know that. Thanks.

PR #364, opened 2020-10-08
Changed Title logo to transparent logo 

Vishv07:
It seems a very small change, but it's important. That's why I didn't raise the issue and Opened a PR.

when our app tab isn't active in a browser, this is how it looked like.
titile

That white background (behind the logo) was not looking good, its looks unprofessional 

now its looks like this 
Changed Title


daltonfury42:
@Vishv07 Agreed, I was going to change this myself in the weekend. 

I am a bit confused because you have other changes from your old PR in here. Can you merge our latest master once into your branch?

Vishv07:
Okay, let me do that first

Vishv07:
@daltonfury42  Okay I fetch all the changes and merged into my branch

maaverik:
Thanks for taking the initiative @Vishv07. I hadn't realized the Simple-Q image doesn't have a background color. We actually don't need two variations of the logo because of this. The whole reason I had suggested that was because of the different background colors and I thought this image had the light blue background. @daltonfury42, what do you think about removing the  LogoLight and LogoDark images and using this image everywhere? 

The only minor problem I see is that that on dark-colored tabs, the logo is not very noticeable, but I think the best solution would be to make a variation of this image with white inside the logo for just tab icons if this seems like a problem. For now, this is perfect.

Screenshot from 2020-10-08 14-53-42

maaverik (approved):

Vishv07:
@maaverik Yeh, we can create a separate new image for the title. but I think this won't be an issue. This is better as of now
Thanks

Issue #365, opened 2020-10-08
Join Queue not working


sajmalyousef:
**Describe the bug**
Join Queue button does not work in the join queue page

**To Reproduce**
1. Go to https://simplq.me/j/AbcQueue
2. Click on Join Queue.'
3. not able to join

**Expected behavior**
User should be able to join AbcQueue

**Screenshots**

![image](https://user-images.githubusercontent.com/13188906/95425853-8e133300-0962-11eb-86b9-3826320fa0b9.png)

maaverik:
This is very interesting. Enter press works all the time even though enter press and the button click trigger the same function. Also, the button click when I have the dev tools window open and if not, the button click doesn't do anything.

sajmalyousef:
Why is the button not clickable?

maaverik:
It's a styling issue. Another div covers the button, that's why the dev tools window fixes it. I never thought something like this could occur.

![Screenshot from 2020-10-08 22-10-50](https://user-images.githubusercontent.com/12583528/95488476-3b159c00-09b3-11eb-9ead-8a770981aedb.png)

I'll check this sometime soon.

maaverik:
I think a z-index might be the easiest fix. @daltonfury42 what do you think?

maaverik:
#366

sajmalyousef:
This has been fixed

maaverik:
Check if #366 looks okay. I think there aren't any more problems.

PR #366, opened 2020-10-08
Bug #365 - join button

maaverik:
I think the issue was because of an unneeded position:absolute. I didn't notice any problems after removing it, but please pull this change and check if there are any side effects.


simplq/src/styles/homePage.module.scss
```diff
@@ -300,6 +302,7 @@
       top: 0;
       left: 0;
       background: rgba(0, 0, 0, 0.5);
+      z-index: 10;
     }
   }
 }
```
maaverik, line 308: This is not supposed to be here. Something's messed up.

simplq/src/styles/joinPage.module.scss
```diff
@@ -1,70 +1,69 @@
...
-	display: flex;
-	flex-wrap: wrap;
-	.simplq-logo {
-		padding-top: 1.1rem;
-		z-index: 3;
-	}
-	.queue {
-		margin: auto;
-		position: absolute;
-		top: 1rem;
```
maaverik, line 30: Removing this position: absolute is my only change.

simplq/src/styles/homePage.module.scss
maaverik: Fixed this.

simplq/src/styles/joinPage.module.scss
Vishv07: keep all css as it is , just add 2 lines of css in .form class
position : relative;
z-index:10;

maaverik: This would probably work too, but I felt like the position: absolute wasn't needed. I didn't notice any problems after removing it. What was your reason for adding it?

Vishv07: See here, I think that was needed, without that queue name isn't perfectly centered.
image

maaverik: Okay, I see. I'll change to the z-index then.

Vishv07:
Done! Works fine

daltonfury42:
I just checked, and the footer links and button in not clickable

maaverik:
All but one button on the footer works for me. This is again because of the position:absolute. The queue div covers parts the footer. Let me try fixing this using a z-index. The better approach would be doing away with position:absolute, but still centering the div. I'll see which one works best.

maaverik:
Can you check if everything's fine now @daltonfury42? I don't notice any problems with my screen size at least. I went with z-index instead of removing the absolute position and using a padding since it messes up exact centering on mobile views. If you can think of a better approach we can try it, else we can go with this.

daltonfury42 (approved):

PR #371, opened 2020-10-10
Copy button expanding on clicking #303

sofistoyanova:


daltonfury42:
I think height wise we are good. We need to give the entire button a fixed with. I'm thinking `min-width`, and in `rem`s. Can you try that?

maaverik:
In addition to the above comments, please move these styles to the adminpage.module.scss file and import them here like 
`import styles from '../../../styles/adminPage.module.scss';`
Take a look at the other component files in the app for reference. We like to keep all the styling for specific pages in the respective scss files.

PR #373, opened 2020-10-10
Fixing memory leak caused by setTimeout

arc30:
setTimeout should be cleared when a component unmounts.

### Description of change

Used useEffect hook's cleanup mechanism to call clearTimeout


daltonfury42 (approved):
Nice!

maaverik:
Good catch. Extra points for adding the link with detailed info for novices like me :).

PR #375, opened 2020-10-11
button width issue fixed

mubashiralisiddiqui:
button width issue fixed


mubashiralisiddiqui:
Css issue fixed

daltonfury42:
@mubashiralisiddiqui This would increase the width for all buttons globally. Can you instead apply this to only the particular button? I'm thinking we should give the button an id and define the widths and height in admin css file

daltonfury42:
Stale PR, closing. Feel free to reopen

PR #382, opened 2020-10-17
Footer Margin Spacing and CSS

daltonfury42:
Fix footer margin, standardising like on home page


maaverik:
To me, the logo is too small and the button too big. That's the only thing I don't personally like.

maaverik (approved):

daltonfury42:
Agreed, I was lazy to fix those... I'm too engrossed with Auth stuff now. Can you take a look / raise an issue?

PR #383, opened 2020-10-17
Clickable logo Logic Enhancement

navaneeth-spotnana:
1. If a onCLick handler is passed, it should be given preference.
2. Added a default when initialising


simplq/src/components/common/ClickableLogo.jsx
```diff
@@ -4,12 +4,17 @@ import React from 'react';
...
 export default (props) => {
-  let onClick = {};
-  if (props.history) {
-    onClick = () => props.history.push('/');
-  } else if (props.onClick) {
+  let onClick = () => {
+    window.location = '/';
+  };
+  if (props.onClick) {
+    // if a handler provided, use it
```
maaverik, line 11: We could use `import { useHistory } from 'react-router-dom';` and remove passing history altogether. That would be cleaner.

daltonfury42: I see. Let me check.

daltonfury42: Done. Using useHistory was neater. On using it, it goes. to the previous scroll state of the home page, which looked odd. Hence I had to make sure that we smooth scroll to top on every mount.

It seems to be working well, we can fix if issues come later.

maaverik (approved):

PR #386, opened 2020-10-18
Add Auth JS Code

daltonfury42:
This is the code to get the JWT token, and send it with each backend call.

No more amplify. We don't also authorise queues created by non-signed in users. So anyone with the admin url will be able to manage the queue. These become sort of public queues.

We should think if/how we should protect them. Also, when someone signs up after creating a public queue, we should migrate his queues to private.  In follow up PRs.

Backend change: https://github.com/SimplQ/simplQ-backend/pull/95


simplq/src/components/ErrorHandler.jsx
```diff
@@ -25,7 +25,7 @@ export class ErrorBoundary extends React.Component {
     // You can also log the error to an error reporting service
     // logErrorToMyService(error, errorInfo);
     console.log('In componentDidCatch');
-    useDispatch(setErrorNotifOpen(true));
+    store.dispatch(setErrorNotifOpen(true));
```
daltonfury42, line 28: hook not allowed in class component

simplq/src/store/appSlice.js
```diff
@@ -14,7 +14,7 @@ const appSlice = createSlice({
   initialState: {
     errorText: '',
     notificationPermission: getNotificationStatus(),
-    loggedInUser: null,
+    isLoggedIn: false,
```
daltonfury42, line 17: We can't set this object in the store, as it contains functions inside it, which mutates itself, an redux doesn't like that.

maaverik (approved):

PR #388, opened 2020-10-19
social icons added

Vishv07:
First of all sorry for the late PR. I was supposed to do very earlier, But my health was not well so that's why
but here it is.


daltonfury42:
Mahn... always health first, this can wait till you have completely recovered.

simplq/public/index.html
```diff
@@ -26,6 +26,7 @@
       rel="stylesheet"
     />
     <link href="https://fonts.googleapis.com/css?family=Pacifico" rel="stylesheet" />
+   <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.15.1/css/all.min.css" integrity="sha512-+4zCK9k+qNFUR5X+cKL9EIR+ZOhtIloNl9GIKS57V1MyNsYpYcUrUeQc9vNfzsWfV28IaLL3i96P9sdNyeRssA==" crossorigin="anonymous" />
```
daltonfury42, line 29: Why did we need this? We are pulling icons from Material UI, right?

simplq/src/styles/footer.module.scss
```diff
@@ -53,12 +53,19 @@
         & > ul {
           margin: 0;
           padding: 0;
+          display: inline-flex;
           list-style-type: none;
-
+          li + li{
+            padding-left:1rem;
+          }
```
maaverik, line 60: Is li + li a typo or does it do something?

simplq/public/index.html
Vishv07: That's Right, But Material UI doesn't have an icon for MEDIUM. So for that, I have used font-awesome CDN

simplq/src/styles/footer.module.scss
Vishv07: Of course, it does something. For spacing between those icons, I have used this technique, What it does is. It adds padding in all <li> except first one.

simplq/public/index.html
daltonfury42: Got it. Is it possible to download the vector image from font awesome or somewhere else and keep it in `public/images` and use it?

Rationale is that we don't want clients to download the entire awesome fonts assets, what do you think?

Vishv07: Yes, this can be done too. Let me know if that's the way we are following. So I can make it done

daltonfury42: Yes, let do this. Download the asset and use it. We don't want to have two icon sets. Material UI icons also, I am planning to optimise later on.

simplq/src/styles/footer.module.scss
Vishv07: @daltonfury42 check now

simplq/src/components/common/Footer.jsx
```diff
@@ -44,24 +49,34 @@ export default () => (
...
             <li>
               <a href="https://medium.com/@raimazach/virtualizing-queues-a-long-overdue-alternative-to-physical-queues-bfdc4b51070f">
-                Medium
+               <img src="images/ICON-MEDIUM.jpg" />
               </a>
             </li>
             <li>
-              <a href="https://www.youtube.com/channel/UCAb9PSXvrGZ4vvSneK1Nrow">Youtube</a>
+              <a href="https://www.youtube.com/channel/UCAb9PSXvrGZ4vvSneK1Nrow">
+                <YouTubeIcon className={styles['youtube-icon']} fontSize={"large"}/>
```
daltonfury42, line 68: Why did we have to add this extra class just for youtube icon?

daltonfury42 (approved):

simplq/src/components/common/Footer.jsx
Vishv07: Bcz that's youtube icon looks way too small than others in large size. So In order to maintain size, I have used bit CSS there.

daltonfury42: Got it. Can you add a small one line comment here:
```
{/* Youtube Icon was looking smaller that others, setting heigh to match other icons */}
```

Vishv07: @daltonfury42 Done!

daltonfury42: Thaaanks

daltonfury42:
I'm working currently on the Auth side... https://github.com/SimplQ/simplQ-frontend/issues/389 will be a nice thing to look into, if you are interested.

Vishv07:
@daltonfury42 Sure, I'll have a look

PR #392, opened 2020-10-26
My Queues on Homepage

daltonfury42:
Add my queues to home page for logged in users

image

### DIFF
--- a/simplq/src/components/common/LoginButton.jsx
+++ b/simplq/src/components/common/LoginButton.jsx
@@ -1,6 +1,6 @@
 import React, { useState } from 'react';
 import { GoogleLogin, GoogleLogout } from 'react-google-login';
-
+import { useHistory } from 'react-router';
 import { useDispatch, useSelector } from 'react-redux';
 import { Avatar, Button } from '@material-ui/core';
 import { setErrorNotifOpen } from '../../store/appSlice';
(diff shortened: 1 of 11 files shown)


simplq/src/components/common/LoginButton.jsx
```diff
@@ -26,6 +29,7 @@ const LoginButton = () => {
 
   const onLogoutCallback = () => {
     Auth.logOut();
+    history.push('/');
```
daltonfury42, line 32: Take the user back to homepage on logout. Otherwise, in the admin page, the unauthenticated user might not have permission, he will get Unauthorised error and will be confused.

simplq/src/components/common/Nav/LeftNav.jsx
```diff
@@ -21,9 +21,6 @@ const LeftNav = ({ open, toggleClose }) => {
             How it works
           </a>
         </li>
-        <li>
-          <a>My Queues</a>
-        </li>
         <li>
```
maaverik, line 24: Instead of removing this, what do you think about making it go to the homepage where the queues are listed, i.e., redirecting then scrolling the queues section?

maaverik (approved):

maaverik:
The plan is to restrict a user to 3 queues at a time, right? Is that restriction set up in the backend already with a descriptive error message?

simplq/src/components/common/Nav/LeftNav.jsx
daltonfury42: Done. I have made that in a follow up PR.

daltonfury42:
> The plan is to restrict a user to 3 queues at a time, right? Is that restriction set up in the backend already with a descriptive error message?

Not yet.

PR #393, opened 2020-10-26
Extend Home Navbar to Admin Page

daltonfury42:


simplq/src/styles/navbar.module.scss
```diff
@@ -59,6 +59,8 @@
   // Login Button
   li:last-child {
     position: fixed;
+    button {background-color: white;}
```
daltonfury42: The logout button at the bottom didn't have a background color, so it was not visible. Temproarily setting it to while, we can design and tweak this later:

image

simplq/src/components/common/Nav/LeftNav.jsx
```diff
@@ -16,13 +16,22 @@ const LeftNav = ({ open, toggleClose }) => {
   return (
     <div>
       <ul className={styles['left-nav']} open={open}>
+        {page === 'home' ? (
+          <li>
+            <a tabIndex={0} role="link" onKeyDown={scrollToHowItWorks} onClick={scrollToHowItWorks}>
+              How it works
+            </a>
+          </li>
+        ) : null}
```
daltonfury42, line 25: Displaying this only on home page for now, as the scroll doesn't work as it's not on home page.

simplq/src/components/common/Nav/LeftNav.jsx
```diff
@@ -16,13 +16,22 @@ const LeftNav = ({ open, toggleClose }) => {
...
+        {page === 'admin' ? (
+          <li>
+            <a tabIndex={0} href="/">
+              My Queues
+            </a>
+          </li>
+        ) : null}
```
maaverik, line 32: Same question as mentioned in the other PR.

maaverik (approved):

simplq/src/components/common/Nav/LeftNav.jsx
daltonfury42: My Queues is shown on admin page navbar only. And it's just a link to "/". This is a naive implementation.

I didn't want to put it on home navbar as well, because it won't do anything, and then the user would think that the link is not working.

Issue #320, opened 2020-10-04
[Admin dashboard]Add member button renaming


sajmalyousef:
Rename the join queue button to Add since this is within the admin dashboard  and  is misleading
Change text 
![image](https://user-images.githubusercontent.com/13188906/95010111-67908780-0644-11eb-93f3-d09391133f0d.png)

daltonfury42:
I was also thinking of this.. This same text would also come in the join queue page, so we should do something neutral.

"Add", "Add to Queue" all works.

sajmalyousef:
Add to Queue  is like stating the obvious? and longer ?

daltonfury42:
Stating the obvious is fine? Long two word buttons == bigger buttons? That was my thought process, but anything is fine

arkumari2000:
Can I do this if this is open and Add to Queue or Add is it?

maaverik:
Hi @arkumari2000, a PR is already open for this pending some changes. Please take a look at the other open issues and see if there's something you'd like to work on.

PR #394, opened 2020-10-26
Add member button renaming

maaverik:
#320


simplq/src/components/pages/Join/Form.jsx
```diff
@@ -68,9 +68,16 @@ export function JoinQueueForm(props) {
     });
   };
 
-  const JoinButton = () => (
-    <div>{addingInProgress ? <LoadingIndicator /> : <JoinQButton onClick={handleClick} />}</div>
-  );
+  const ActionButton = () => {
```
daltonfury42, line 71: we have to pass props here.

simplq/src/components/pages/Join/Form.jsx
```diff
@@ -98,7 +105,7 @@ export function JoinQueueForm(props) {
         onChange={handleContactChange}
         onKeyDown={(e) => handleEnterPress(e, handleClick)}
       />
-      <JoinButton />
+      <ActionButton />
```
daltonfury42, line 108: I'd prefer calling this "CreateTokenButton"

simplq/src/components/pages/Admin/AddMember.jsx
```diff
@@ -17,6 +17,7 @@ export default (props) => {
           buttonName="Add"
```
daltonfury42, line 17: Can we solution this around `buttonName` rather than going this route?

maaverik: Okay

daltonfury42: coz I think that will be a safer solution.

simplq/src/components/pages/Join/Form.jsx
maaverik: It's using the JoinQueueForm's props here, so no need.

simplq/src/components/pages/Join/Form.jsx
maaverik: Sure

simplq/src/components/pages/Join/Form.jsx
```diff
@@ -68,14 +68,14 @@ export function JoinQueueForm(props) {
...
-  const ActionButton = () => {
+  const CreateTokenButton = () => {
     if (addingInProgress)
       return (
         <div>
           <LoadingIndicator />
         </div>
       );
-    if (props.page === 'admin') return <AddToQueueButton onClick={handleClick} />;
+    if (props.buttonName === 'Add') return <AddToQueueButton onClick={handleClick} />;
```
daltonfury42, line 78: We can pass buttonName down to `CreateTokenButton`, then we won't need two separate `JoinQueueButton` and `AddToQueueButton `

maaverik: Done

simplq/src/components/pages/Join/Form.jsx
```diff
@@ -75,7 +75,7 @@ export function JoinQueueForm(props) {
           <LoadingIndicator />
         </div>
       );
-    if (props.buttonName === 'Add') return <AddToQueueButton onClick={handleClick} />;
+    if (props.buttonName === 'Add') return <Button onClick={handleClick}>Add to Queue</Button>;
```
daltonfury42, line 78: what I meant was,
```
return <JoinQueueButton onClick={handleClick}>{props.buttonName}</Button>
```

no need of an if/else anywhere.

maaverik: I'm not keen on this idea. Then the whole joinQueueButton is just a pointless wrapper.

daltonfury42: ok. then lets remove that. `return <Button onClick={handleClick}>{props.buttonName}</Button>;`

I want to avoid this `if (props.buttonName === 'Add') ... else ...` thing

daltonfury42: can we then remove any other unwanted button wrappers and use Button/StandardButton in it's place?

maaverik: I've removed the joinQueueButton since it's not used much anymore. We'll remove the other as required.

daltonfury42 (approved):
Yo. This buttons can definitely be handled in better ways... we can refacto later.

daltonfury42:
Also, I don't use the storybook UI for developing these... I've not used that UI for a long time, hence that's also something not adding a value, and another candidate for removal.

Storybook was my idea, we realise that it's not needed, and we can remove it.

PR #395, opened 2020-10-26
Delete Queue

daltonfury42:
Delete queue impl


simplq/src/components/pages/Home/MyQueues.jsx
```diff
@@ -2,20 +2,32 @@ import React from 'react';
...
-  if (myQueues.length === 0) {
-    return null;
-  }
+  const onDeleteClick = (queueId) => {
+    deleteQueue(queueId)
+      .then(() => getMyQueues().then((myNewQueues) => dispatch(setMyQueues(myNewQueues))))
+      .catch((err) => handleApiErrors(err))
+      .then(() => history.push('/'))
+      .catch((err) => handleApiErrors(err));
+  };
```
maaverik, line 23: Moving this to utilFns would be a good idea.

daltonfury42: In a PR, we can remove all `.catch((err) => handleApiErrors(err))` globally, and use a axios interceptor to handle this in one place.

maaverik: I actually meant moving the onDeleteClick fn since it's used in MyQueues and DeleteQueues

maaverik: As for the interceptor, I'll look at the error boundary and interceptor probably this weekend.

daltonfury42: Done. I have moved this into the queue service.

simplq/src/services/queue.js
```diff
@@ -29,3 +30,10 @@ export const getStatusByName = async (queueName) => {
   const response = await makeRequest('get', `/queue/status?queueName=${queueName}`);
   return response.data;
 };
+
+export const deleteQueue = async (queueId) => {
+  const response = await makeRequest('delete', `/queue/${queueId}`);
+  await getMyQueues().then((myQueues) => store.dispatch(setMyQueues(myQueues)));
+  history.push('/');
+  return response.data;
```
maaverik, line 38: Did you miss adding the catch or should I add exception handling while adding the axios interceptor?

maaverik (approved):

simplq/src/services/queue.js
daltonfury42: Let's add in one place... if you are planning to do that soon

PR #399, opened 2020-10-30
Delete queue

daltonfury42:
In the previous delete queue PR (#395) I had tried to make the router history available outside react, by following [this](https://stackoverflow.com/questions/44153517/how-to-access-history-object-outside-of-a-react-component). But for some reason, I am not able to get that to work. So I had reverted that PR a few ways ago, now re-raising that PR without the history change.


simplq/src/components/Routes.jsx
```diff
@@ -12,14 +12,16 @@ export default () => {
...
+        <Router>
+          <Switch>
+            <Route path="/" exact component={Home} />
+            <Route path="/queue/:queueId" exact component={AdminQueue} />
+            <Route path="/j/:queueName" exact component={JoinQueueWithDetails} />
+            <Route path="/token/:tokenId" exact component={QueueStatus} />
+            <Route path="/privacy" exact component={TermsOfService} />
+            <Route component={PageNotFound} />
+          </Switch>
+        </Router>
```
daltonfury42, line 24: Moved router into the Routes file

simplq/src/components/pages/Home/CreateJoinForm.jsx
```diff
@@ -7,10 +8,11 @@ import { handleEnterPress, isQueueNameValid } from '../../common/utilFns';
 import InputField from '../../common/InputField';
 import LoadingIndicator from '../../common/LoadingIndicator';
 
-const CreateJoinForm = ({ history }) => {
+const CreateJoinForm = () => {
   const [textFieldValue, setTextFieldValue] = useState('');
   const [invalidMsg, setInvalidMsg] = useState('');
   const [createInProgress, setCreateInProgress] = useState(false);
+  const history = useHistory();
```
daltonfury42, line 15: We didn't have to pass down history, using the hook here.

maaverik:
If someone deletes their queue and goes back using the browser back button to the admin page again, they see the old admin page as it was before. Can we at least show a message in the main admin container saying 'Your queue was deleted' or something like that?

daltonfury42:
ah yes we should do this. I'll create an issue so that we don't forget this

maaverik (approved):

PR #404, opened 2020-10-31
More Admin Layout Improvements - For mobile screens

daltonfury42:
Social Media icons were not wrapping properly, they were going off screen:
image


simplq/src/components/pages/Admin/Token.jsx
```diff
@@ -105,7 +105,7 @@ function Token(props) {
       </div>
       <div className={styles['token-details']}>
         <div className={styles['token-name-time']}>
-          <p>Join time 11:02 pm</p>
```
daltonfury42, line 108: Removing joining time as it's taking up more space... but this I was not able to completely solve still... it's not very straight forward:

image


We need to flex wrap properly or do something on small screens...

maaverik:
It looks like this messes up the centering on mobile view which is present on the main site.

Screenshot from 2020-11-01 16-28-08

daltonfury42:
got it.. can I fix this as a follow up PR?

maaverik:
Okay

maaverik (approved):

PR #405, opened 2020-10-31
Fix Join/Status Page Margins and Footer 

daltonfury42:
1. Fix footer to bottom
2. Add Margins to Status and Join Page, fix Header section styles
3. Demove login button for now

### DIFF
--- a/simplq/src/components/Layout.jsx
+++ b/simplq/src/components/Layout.jsx
@@ -1,13 +1,18 @@
 import React from 'react';
 import Routes from './Routes';
 import Footer from './common/Footer';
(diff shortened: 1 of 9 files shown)


maaverik:
This somehow messed up some background colors in case the default background of the browser is not white. I'm not sure where the problem is.

Screenshot from 2020-11-01 16-10-40
Screenshot from 2020-11-01 16-10-23

maaverik:
I've removed that global background:white setting. So this background color problem should be fine now.

maaverik (approved):

PR #407, opened 2020-11-01
Adding info and error popup notifications

maaverik:
Adding info popups for leave and delete queue operations similar to error popups.


simplq/src/components/ErrorHandler.jsx
```diff
@@ -25,7 +22,7 @@ export class ErrorBoundary extends React.Component {
     // You can also log the error to an error reporting service
     // logErrorToMyService(error, errorInfo);
     console.log('In componentDidCatch');
-    store.dispatch(setErrorNotifOpen(true));
+    store.dispatch(setErrorPopupMessage(true));
```
daltonfury42, line 25: Wanted to understand, why do we set this to "true" here?

daltonfury42: We are not using `error` and `errorInfo` here? Or maybe these should not be a popul, we should let them go down and be caught by sentry?

daltonfury42 (commented):
This is a nice one.

daltonfury42 (approved):

simplq/src/components/ErrorHandler.jsx
maaverik: What I remember from when I wrote this is that the code never reaches this point. That's probably why I didn't notice this was set to true. I'll try to figure this out soon.

maaverik: Do you want to merge this now and use this also for the prod branch?

PR #417, opened 2020-11-05
Add scroll animations to home page

daltonfury42:
A quick experiment! 

Inspired by https://absurd.design
Powered by https://michalsnik.github.io/aos/

Preview: https://animate.dpmhbfuiok7f8.amplifyapp.com/

### DIFF
--- a/simplq/package-lock.json
+++ b/simplq/package-lock.json
@@ -5063,6 +5063,16 @@
         "normalize-path": "^2.1.1"
       }
     },
+    "aos": {
+      "version": "2.3.4",
+      "resolved": "https://registry.npmjs.org/aos/-/aos-2.3.4.tgz",
+      "integrity": "sha512-zh/ahtR2yME4I51z8IttIt4lC1Nw0ktsFtmeDzID1m9naJnWXhCoARaCgNOGXb5CLy3zm+wqmRAEgMYB5E2HUw==",
+      "requires": {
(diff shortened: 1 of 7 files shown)


maaverik:
This is a nice addition. I noticed a few parts that are not very smooth though. In the Why SimplQ section, the first two images always fade in together while the third one comes later. Could you make all three fade in together or make them fade in one after the other independently?  The create queue section below it looks great, but the join queue section fades in a little late than I would like. The create queue section fades in when I almost scroll to the middle of the section while the join queue section waits till I reach the bottom of the section to fade in, which I personally don't like. Could you try making it fade in like the create queue section?

daltonfury42:
I had been experimenting with different approaches in the different sections. 

1. Why SimplQ section: Independent for each card. 
2. How it works section: Fade up, all together, overrode a deply `data-aos-duration="800"`
3. How to join: Independent, coming from each direction

For 1, (Why simplQ), agreed, zoom in all at once might would be better. Let me know
For 3 (How to join) I think this is happening because it's three independent things fading in... not all at one go. This might be difficult to get around.

daltonfury42:
I don't want both create queue and  join queue to have the same variation, each adjacent section animations should be different.

maaverik:
The Why simplQ part is much better now. I still have mixed feelings about the join simplQ part taking so long to fade in but it's not a big deal.

maaverik (approved):

PR #418, opened 2020-11-07
Avoid Movement of Other Components on Expansion

daltonfury42:
The footer and the main content keeps moving up and down when you expand and collapse:

ezgif com-gif-maker

Avoiding this by fixing the footer down below visible screen, and also not centring the text of empty queue. I tried to avoid the latter, but couldn't make both work.

### DIFF
--- a/simplq/src/components/pages/Admin/index.jsx
+++ b/simplq/src/components/pages/Admin/index.jsx
@@ -93,17 +93,13 @@ export default (props) => {
   );
 
   return (
-    <>
+    <div className={styles['admin-content']}>
       <AdminNavbar />
       <HeaderSection />
       <div className={styles['main-body']}>
-        <div
-          className={tokens?.length > 0 ? styles['token-list-with-content'] : styles['token-list']}
-        >
-          <TokenList tokens={tokens} queueId={queueId} removeTokenHandler={removeToken} />
-        </div>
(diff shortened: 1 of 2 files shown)


maaverik:
I'm pretty sure this problem wasn't there last week. What changes did we make to the admin page this week? One of them might have caused this.

maaverik:
I think there used to be more space between the footer and the sidepanel so the expansion used to fill up that space and not move the entire page.

daltonfury42:
I checked out 4 weeks old code https://github.com/SimplQ/simplQ-frontend/commit/00e750cf0a108054237d7fde8c8108029f20a6af and the problem is there.

daltonfury42:
Actually the diff is mostly whitespace changes because I wrapped a lot of css inside one more nesting... The main change is only https://github.com/SimplQ/simplQ-frontend/pull/418/commits/e296a9956ad5f13c63737bf027ec27f00ba884f7 and https://github.com/SimplQ/simplQ-frontend/pull/418/commits/ec246d9dd9a84ce9c00df12cc59e56675273f033

Hide whitespace when you review the diff:
image

maaverik (approved):

maaverik:
I thought I'd approved already. This hide whitespaces setting is useful :).

PR #429, opened 2020-12-13
Handle API Failures at One Place

daltonfury42:
We move up handling API error so that it would be called on all API errors without the caller having to explicitly call.

Few more tiny optimisations:
1. If no errors, we resolve `response.data` also in a single place.


simplq/src/services/queue.js
```diff
@@ -3,35 +3,31 @@ import { store } from '../store';
 import { setMyQueues } from '../store/appSlice';
 
 export const getMyQueues = async () => {
-  const response = await makeRequest('get', '/queues');
-  return response.data.queues;
```
daltonfury42, line 7: Tiny Fix: This was resolving to `.queues` which was an outler.

maaverik (approved):
Nicely done. I didn't check every change, but there are no noticeable mistakes. We can fix later if we find any.

PR #435, opened 2020-12-16
#421 - Add primary to copy to clipboard button

ahsanbagwan:
Height of the button goes wonky


ahsanbagwan:
@daltonfury42

daltonfury42:
Thanks for the PR. I was looking at our code, and saw that an unwanted div was causing the issue for you. 

Digging deeper, I feel that our current deign is not that great. I did a huge refactor, and while doing it, #421 got resolved on its own.

daltonfury42:
Refactor PR: https://github.com/SimplQ/simplQ-frontend/pull/436

Issue #277, opened 2020-10-01
[Refactoring][EngDebt] Revisit Storybook


daltonfury42:
We had decided initially to use storybook.js to write reusable components and also fixed on scss for CSS. While using SCSS has turned out to be a right choice, I feel storybook doesn't give us much value addition. It is suitable for larger projects.

We could create a reusable button, but then, that could have been easily done by reusing the css put in a common file.

@maaverik what do you think? Can we remove storybook from our project?

maaverik:
We could. But it wouldn't make much of a difference if we keep it or not. It's a dev dependency I believe so it doesn't slow down the site or anything. And even removing it just means renaming and modifying a few files. I'm fine either way.

daltonfury42:
If we do something, be it a feature, or some framework or some library, and if it's not giving us value, I'd see it as an overhead / eng-debt... I'll remove some day.

It's just adding extra complexity to our codebase which we don't need.

maaverik:
I feel that storybook has its uses. We have more shared components now than before in our common folder and it's starting to become hard to keep track of them. Only a few are currently configured to show up on storybook. I say we integrate all of those with storybook so that new contributors can just run storybook and see all the common components available and choose what they need. The number of common components is only going to go up anyway.

PR #436, opened 2020-12-19
Use StandardButton directly in components

daltonfury42:
1) We were defining the Buttons in a `.stories` file and importing them, this felt like an unwanted indirection.
2) We have two types of buttons, the normal default one (dark) and outlined. Adding an extra option to pass an icon optionally to these buttons. 
3) Also make "Copy to clipboard" primary, using InfoPopup instead of the timer and color change. 
4) Tweaked the CSS a bit for buttons, better spacing and centering.


maaverik:
Yeah, removing actual logic from storybook files is a good idea. I think storybook is just meant to showcase components, and we shouldn't really define new components for reuse within storybook files.

We can actually decide once and for all if we should keep storybook. While writing this old [comment](https://github.com/SimplQ/simplQ-frontend/issues/277#issuecomment-704092266), I had an idea about a way of making use of storybook better. Let me try implementing that and if it still doesn't seem worth it, let's ditch storybook completely.

simplq/src/components/pages/Admin/ShareQueue.jsx
```diff
@@ -1,44 +1,23 @@
...
-    // changes back to old text and color after 5 seconds
-    const timerId = setTimeout(() => setClicked(false), 5000);
-    return () => clearTimeout(timerId);
-  }, [clicked]);
+  const dispatch = useDispatch();
 
   const handleShareButtonClick = () => {
-    setClicked(true);
+    dispatch(setInfoPopupMessage('Copied to clipboard'));
   };
```
maaverik, line 14: We could change this to 'Copied queue link to clipboard' or 'You can now share queue details from the clipboard'

simplq/src/components/common/Button/button.module.scss
```diff
@@ -28,8 +32,10 @@
   }
 }
 
-.standard-button img {
-  height: 1.5rem;
+.icon {
+  display: inline;
+  vertical-align: middle;
+  margin-right: 0.5rem;
 }
```
maaverik, line 40: Could the icon be made horizontally centered?  Compared to the live site, it's placed a little higher up relative to the text.

simplq/src/components/pages/Admin/ShareQueue.jsx
```diff
@@ -1,44 +1,23 @@
...
-    );
-
   return (
     <CopyToClipboard text={shareUrl}>
-      <ShareButton onClick={handleShareButtonClick} outlined={!clicked}>
-        <CopyToClipboardButton />
-      </ShareButton>
+      <StandardButton onClick={handleShareButtonClick} icon={<FileCopyIcon />}>
+        Copy to clipboard
+      </StandardButton>
```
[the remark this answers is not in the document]
daltonfury42: Copy Queue Link makes sense. I think we should keep "Share Queue" for a different use case where we can detect that the browser is mobile, and we put a "Share" button if there is support for it. (It will then open the app to share via, Whatsapp, FB etc). I have not checked if this is possible, though.

simplq/src/components/pages/Admin/ShareQueue.jsx
daltonfury42: Copied queue link to clipboard 👍

simplq/src/components/common/Button/button.module.scss
daltonfury42: ah, the Refresh button has the issue, while the copy button looks fine. I am guessing it's because of presence of characters with 'p' and 'y' with a bellow the line stoke. Let me check.

image

daltonfury42: Does it look fine now?
image

If yes, I think it has to do with the icons, not sure how to fix.

daltonfury42: did you mean vertical or horizontal? Can you give a screenshot?

maaverik: My bad, I meant vertical. But it's as you've shown. I think it can be fixed though.

With these changes, it looks like this.
Screenshot from 2020-12-20 01-55-14


But, on the main site, it is centered.
Screenshot from 2020-12-20 01-55-19

So I thought it might be a spacing issue.

daltonfury42: That was a easy fix, after a good night's sleep. Done

maaverik (approved):

maaverik:
Could you pull the latest code and overwrite the button stuff with your changes?

daltonfury42:
Done

PR #437, opened 2020-12-19
Storybook changes

maaverik:
Added storybook files in common component folders for new users to get an idea of reusable components already present.

Couldn't get the fonts to work properly yet cause of incompatibility with storybook and some webpack features.


maaverik:
@daltonfury42, try 'npm run storybook' and see if this is worth having.

daltonfury42:
We can have this and see how it gets used. We can also host this somewhere like storybook.simplq.me 

Should we avoid importing components from the .stories file into the main component? What's your take on that? 

I think what we define in the common module is generic parameterised versions, and we should not try to define all valiants in the .story file and import into the app. Storybook would be just to have the different reusable components in a library, and .stories are just to demonstrate it's different states.

maaverik:
Yeah, let's avoid any kind of imports from inside .storybook files. Let's keep those files just for presentation while using 'npm run storybook'. Hosting this specifically doesn't really seem worth it to me.

And yeah, no need of variants. We can get merge #436 to this to remove the unnecessary buttons we have.

daltonfury42 (approved):

PR #440, opened 2020-12-24
Fix trivial warnings, resolve promise on login autoload when not logged in

daltonfury42:


simplq/src/components/common/LoginButton/index.jsx
```diff
@@ -40,8 +40,14 @@ const LoginButton = () => {
     setLoadingIndicator(true);
   };
 
+  const onAutoLoadFinished = (successLogin) => {
+    if (!successLogin) {
+      Auth.logOut();
+    }
+  };
+
```
daltonfury42, line 48: The login button on start will automatically try to login. In case login did not happen (as user was not previously logged in), we need to resolve the promise in Auth service, else it will wait for ever. This is done by calling `Auth.logOut()`

maaverik (approved):

Issue #439, opened 2020-12-20
Temporary Queue Warning for Unregistered Users


daltonfury42:
![image](https://user-images.githubusercontent.com/11256376/102719374-baa8db00-4313-11eb-88b4-4c4463595c02.png)

Display a ribbon on admin page if the user has not logged in. The text can be fixed later. 

Use `isLoggedIn()` in [auth.js](https://github.com/SimplQ/simplQ-frontend/blob/master/simplq/src/services/auth.js) to check if user is logged in or not.

[Design Link](https://xd.adobe.com/view/ad1db074-03bf-45b1-537b-98d9d524ec82-db2c/screen/fe137c74-d5e2-4fe1-af85-10877cc560b8/) 

Design show it as above navbar. But keeping it above the navbar might be a bit more complex, so we can display the ribbon below the header section:
![image](https://user-images.githubusercontent.com/11256376/104058213-62317300-5219-11eb-84d6-6c498228eaca.png)

yahsashay:
** want  to work on this issue **
we want only the orange color ribbon, not the text and a cross button

daltonfury42:
Sure go ahead. This is a easy issue that can get you started.

We can have the ribbon, the text as in the design and the close button.
Let’s keep the same text as in the design, it can be modified later.

shubh128:
I would like to work on this issue if it's not already being resolved

daltonfury42:
Yes, you can take this up. @yahsashay has not send in a PR yet.

shubh128:
![image](https://user-images.githubusercontent.com/48080672/103174811-21d20c80-488b-11eb-8288-672cc712db7f.png)
I was able to show a warning with this design. However I'm unable to match the exact design required as I can't figure out the CSS. Let me know if this design works for you or if you can help me with the CSS.
Also I was getting some errrors due to CORS being blocked.

daltonfury42:
Send a PR, this is looking just great. I will review it there.

Tiny change: I guess keeping it on top might not be straightforward.  We can display the ribbon it between the navbar and the violet region (above the title). That way you don't have to position it absolutely.

Can you show me the CORS error, few screenshots would help.

shubh128:
![image](https://user-images.githubusercontent.com/48080672/103193229-d4dd4d00-4901-11eb-9947-9c9ae55ea182.png)
I get this error when opening the queue page without signing in.
[...]

PR #442, opened 2020-12-28
Initialized Temporary Queue Warning

shubh128:
Resolves https://github.com/SimplQ/simplQ-frontend/issues/439

### DIFF
--- a/simplq/src/components/common/Popup/TempQueueWarning.jsx
+++ b/simplq/src/components/common/Popup/TempQueueWarning.jsx
@@ -1,0 +1,57 @@
+import React, { useState, useEffect } from 'react';
+import { makeStyles } from '@material-ui/core/styles';
+import IconButton from '@material-ui/core/IconButton';
+import Collapse from '@material-ui/core/Collapse';
+import CloseIcon from '@material-ui/icons/Close';
+import Grid from '@material-ui/core/Grid';
+import { isLoggedIn } from '../../../services/auth';
+
+const useStyles = makeStyles(() => ({
+  root: {
+    width: '100%',
+    background: '#FF8000',
+    color: '#FFFFFF',
+    top: '0',
+    position: 'absolute',
+    zIndex: '100',
+  },
+}));
+
+export default () => {
+  const [open, setOpen] = useState(true);
+  const [loginStatus, setLoginStatus] = useState(false);
+  const classes = useStyles();
+  useEffect(() => {
+    setLoginStatus(isLoggedIn());
+    return () => setLoginStatus(null);
+  }, []);
+  if (loginStatus !== false) return null;
+  return (
+    <div className={classes.root}>
+      <Collapse in={open}>
+        <Grid container direction="row" justify="space-around" alignItems="center">
+          <Grid item>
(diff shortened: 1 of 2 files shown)


simplq/src/components/common/Popup/TempQueueWarning.jsx
```diff
@@ -0,0 +1,57 @@
...
+  root: {
+    width: '100%',
+    background: '#FF8000',
+    color: '#FFFFFF',
+    top: '0',
+    position: 'absolute',
+    zIndex: '100',
+  },
+}));
+
```
maaverik, line 19: @shubh128, could you move this styling code to an scss file and import it for use. That's how it's done everywhere else, so let's keep things consistent.

maaverik (changes requested):

simplq/src/components/common/Popup/TempQueueWarning.jsx
maaverik: If possible, try to reduce the usage of material UI. This is not a hard requirement, we just don't want to be too tightly coupled with any styling library. If this is too hard to do, leave it for now.

daltonfury42: Let's keep it as a hard requirement here. Except for the `CloseIcon`, can you avoid all other material ui usage?

simplq/src/components/common/Popup/TempQueueWarning.jsx
```diff
@@ -0,0 +1,57 @@
+import React, { useState, useEffect } from 'react';
```
daltonfury42, line 1: Move to a separate folder named `ribbon/` inside `common/` with two files,
```
/common/Ribbon/
  - index.jsx                          // for the component
  - ribbon.module.scss                 // for styles
``` 

You can use [LoginButton](https://github.com/shubh128/simplQ-frontend/tree/master/simplq/src/components/common/LoginButton) for refernce.

simplq/src/components/common/Popup/TempQueueWarning.jsx
```diff
@@ -0,0 +1,57 @@
...
+import { isLoggedIn } from '../../../services/auth';
+
+const useStyles = makeStyles(() => ({
+  root: {
+    width: '100%',
+    background: '#FF8000',
+    color: '#FFFFFF',
+    top: '0',
+    position: 'absolute',
+    zIndex: '100',
```
daltonfury42, line 16: Remove `z-index`, `position: 'absolute'` and `top: 0`.

Can you use flexbox?

simplq/src/components/pages/Admin/index.jsx
```diff
@@ -94,6 +95,7 @@ export default (props) => {
   return (
     <div className={styles['admin-content']}>
       <HeaderSection />
+      <TempQueueWarning />
```
daltonfury42, line 98: Lets call this component "Ribbon"

Rationale: We can reuse it multiple times with different content in future.

daltonfury42:
Hi @shubh128 Have not heard back from you. Will you be finishing this?

PR #445, opened 2021-01-03
Fix Page Refresh Time Token Set to Anonymous

daltonfury42:


simplq/src/components/pages/Home/home.module.scss
```diff
@@ -25,7 +25,8 @@
     font-size: larger;
   }
   .create-join-form {
-    margin: auto; // centering
+    margin: auto; // auto for centering
+    padding-top: 2rem;
```
daltonfury42, line 29: minor spacing correction

maaverik (approved):

PR #454, opened 2021-01-11
Create ribbon queue warning

### LINKED ISSUE #439: Temporary Queue Warning for Unregistered Users

rafaelcsva:
resolves #439

### DIFF
--- a/simplq/src/components/common/Ribbon/index.jsx
+++ b/simplq/src/components/common/Ribbon/index.jsx
@@ -1,0 +1,37 @@
+import React, { useState, useEffect } from 'react';
+import CloseIcon from '@material-ui/icons/Close';
+import { isLoggedIn } from '../../../services/auth';
+import styles from './ribbon.module.scss';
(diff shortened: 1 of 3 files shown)


rafaelcsva:
i've just updated the solution of @shubh128 hope it works well

daltonfury42 (approved):
Thanks for the PR :)

Issue #461, opened 2021-01-13
"Login Failed. Please try again" PreventDefault


abhishekpatel946:
**Describe the bug**

**To Reproduce**
1. Go to 'https://www.simplq.me/'
2. Click on 'Home'
3 See error 'Login Failed. Please try again.'

**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/46045014/104463193-7984ad80-55d7-11eb-971d-b65dd4ef249a.png)

**Desktop (please complete the following information):**
 - OS: [Windows10]
 - Browser [Chrome]
 - Version [87]

**Additional context**
On the Landing page unexpected `Login Failed. Please try again.` or when I click on `Home` the same error prompt opens. I think when I landed or refresh the page the form will be submitted by default. Prevent this kind of uncertain form or click.

daltonfury42:
Thanks @abhishekpatel946 for reporting this. I can reproduce the issue on Incognito.

prabureddy:
@abhishekpatel946 Everything works fine in my system. Clear your cache, site settings and check again.

daltonfury42:
Definitely we need better error here. I've raised #462 once that goes in, we will get more logging. On Incognito login fails due to cookies not being enabled, we will have to figure it out later.


![image](https://user-images.githubusercontent.com/11256376/104497911-0b54e080-5601-11eb-9b71-6804f8e6b883.png)

daltonfury42:
@abhishekpatel946 Can you try reproducing the issue on https://dev.simplq.me?

abhishekpatel946:
**I got the point of the problem.**

> If we want to `Block third-party cookies` or `Block all cookies (not recommended)` in chrome.
> Then the `Login Failed. Cookies are not enabled in current environment`. Prompt is activated.

![image](https://user-images.githubusercontent.com/46045014/104541455-ab355d00-5647-11eb-8220-8d8a7eb7f62a.png)


_but if we `Allow all cookies` in Chrome then the prompt is not active._ 
**Note:-** Same thing happens in the incognito mode in Chrome.

abhishekpatel946:
> Thanks @abhishekpatel946 for reporting this. I can reproduce the issue on Incognito.

sure :+1:

maaverik:
What would be the best way to suppress this error message for us? Should we just pass an extra argument to skip the handleApiErrors call in case this particular situation occurs?

daltonfury42:
[...]

PR #462, opened 2021-01-13
Enable Logging of Login Failures

daltonfury42:
Logging error and send the reason to the user. 

image

To figure out what went wrong in #461


simplq/src/components/common/LoginButton/index.jsx
```diff
@@ -23,9 +24,10 @@ const LoginButton = () => {
     setLoadingIndicator(false);
   };
 
-  const onFailureCallback = () => {
+  const onFailureCallback = (resp) => {
+    Sentry.captureMessage(resp);
     Auth.logOut();
```
maaverik, line 29: Why add sentry here? Do we really need to be notified of every login failure?

maaverik (approved):

simplq/src/components/common/LoginButton/index.jsx
daltonfury42: I don't know for sure. I get what you mean. Maybe we can collect this and aggregate at sentry level. Something like, define an alert if more than 30% of last 10 logins failed, trigger a notification, as it could be something wrong at our end.

maaverik: That's a good idea. Gotta check if Sentry supports it.

PR #474, opened 2021-01-16
Fix Spinner on home page, logo to svg and other fixes

maaverik:


simplq/src/components/common/Footer/index.jsx
```diff
@@ -67,7 +67,7 @@ export default () => (
           </div>
           <div className={styles['list-container-item']}>
             <a href="https://medium.com/@raimazach/virtualizing-queues-a-long-overdue-alternative-to-physical-queues-bfdc4b51070f">
-              <img alt="medium-link" src="/images/ICON-MEDIUM.jpg" />
+              <img alt="medium-link" src="/images/ICON-MEDIUM.svg" />
             </a>
```
maaverik, line 71: The old icon had a slight color mismatch (white text instead of changing with background)

simplq/src/components/common/Popup/QrCode.jsx
```diff
@@ -5,17 +5,22 @@ import PrintIcon from '@material-ui/icons/Print';
...
       <h2>Scan this QR to begin!</h2>
       <QRCode value={url} />
-      <p style={{ textAlign: 'center' }}>{`or visit ${url}`}</p>
+      <p style={{ textAlign: 'center' }}>
+        {'or visit '}
+        <a href={url} target="_blank" rel="noopener noreferrer">
+          {url}
+        </a>
+      </p>
     </div>
```
maaverik, line 24: This is not really necessary, but I figure if the queue owner shares the QR PDF, users could join easier if it contains a link already.

daltonfury42: I think we should do this. It also give the look and feel of a link

simplq/src/components/pages/Home/CreateJoinForm.jsx
```diff
@@ -57,16 +57,18 @@ const CreateJoinForm = () => {
...
+          <>
+            <div>
+              <StandardButton onClick={handleCreateClick}>Create Queue</StandardButton>
+            </div>
+            <div>
+              <StandardButton onClick={handleJoinClick}>Join Queue</StandardButton>
+            </div>
+          </>
+        )}
       </div>
```
maaverik, line 72: Instead of the spinner overlapping with the join button which was pretty ugly, moved the code around a bit. Now there won't be the overlap between the spinner and button. There will still be a style change while the spinner is active because the enclosing container is flexed, so when the buttons aren't there anymore, it shrinks accordingly. I can't think of an elegant fix for this so I'll raise an issue for now.

simplq/src/components/pages/Admin/index.jsx
```diff
@@ -86,7 +86,7 @@ export default (props) => {
       <div className={styles['main-button-group']}>
         <div className={styles['admin-button']}>
           <StandardButton onClick={generateQrCOde} icon={<CropFreeIcon />} outlined>
-            Generate QrCode
+            Generate QR Code
```
daltonfury42, line 89: Do we want "Generate QR Code" or "Show QR Code"?

simplq/src/components/pages/Home/CreateJoinForm.jsx
```diff
@@ -57,16 +57,18 @@ const CreateJoinForm = () => {
...
+        ) : (
+          <>
+            <div>
+              <StandardButton onClick={handleCreateClick}>Create Queue</StandardButton>
+            </div>
+            <div>
+              <StandardButton onClick={handleJoinClick}>Join Queue</StandardButton>
+            </div>
+          </>
+        )}
```
daltonfury42, line 71: This makes sense

daltonfury42 (approved):

daltonfury42:
Let's be a bit more descriptive on PR names and especially commit messages. This could be something like,

"Fix Spinner on home page, logo to svg and other fixes"

simplq/src/components/pages/Admin/index.jsx
maaverik: I personally prefer generate.

daltonfury42: Cool. 👍

PR #478, opened 2021-01-19
Fix spinners modifying page styles when active

mathkruger:


simplq/src/components/common/LoadingIndicator/loading.module.scss
```diff
@@ -3,5 +3,5 @@
 .main {
   @include center-vertically();
   @include center-horizontally();
-  min-height: 100vh;
+  min-height: 30vh;
```
daltonfury42, line 6: This change makes sense. 

Now I'm wondering if we can completely remove this min-height here. @mathkruger What do you think?

mathkruger: I tested without the "min-height" and, on some pages, it didn't look any good. I tried some values and it worked on every page where I could see this component.

maaverik (approved):
This looks good. This min-height value doesn't really make the spinner look bad in other places, so it's fine. Thanks for the PR @mathkruger.

PR #482, opened 2021-01-24
Queue stats

### LINKED ISSUE #424: Show Queue Stats On Queue Joining Page
It will be good to show the total number of people in the queue, before the user joins the queue. We could show the entire queue stats in the joining page.

mradenovic:
Extract Queue Stats and display stats where appropriate.

Resolves #424


simplq/src/components/pages/Status/QueueDetails.jsx
```diff
@@ -40,18 +23,7 @@ export default (props) => {
...
-          value={queueStatusResponse?.numberOfActiveTokens}
-          large
-        />
-        <DetailRow title="Queue creation time:" value={creationTime} />
-        <DetailRow
-          title="Total number of people joined in queue:"
-          value={queueStatusResponse?.totalNumberOfTokens}
-        />
-      </div>
+      <QueueStats queueStatus={queueStatusResponse} />
```
maaverik, line 26: Since you moved this component out, please also remove the styles specific to QueueDetails in status.module.scss

simplq/src/components/common/QueueStats/queueStats.module.scss
```diff
@@ -0,0 +1,23 @@
...
+
+.detail {
+  display:flex;
+  flex-direction: column;
+  padding: 0.5rem 2rem;
+
+  .detail-row {
+    display: flex;
+    align-items: center;
+    height: 24px;
```
daltonfury42, line 11: nit: Can you covert to rem here?

simplq/src/components/common/QueueStats/queueStats.module.scss
```diff
@@ -0,0 +1,23 @@
+@import '../../common.scss';
+
+.detail {
+  display:flex;
+  flex-direction: column;
+  padding: 0.5rem 2rem;
+
+  .detail-row {
+    display: flex;
+    align-items: center;
```
daltonfury42, line 10: is this required? Let's do a `justify-content: space-between;` here

simplq/src/components/common/QueueStats/queueStats.module.scss
```diff
@@ -0,0 +1,23 @@
+@import '../../common.scss';
+
+.detail {
+  display:flex;
+  flex-direction: column;
+  padding: 0.5rem 2rem;
```
daltonfury42, line 6: Can you add a `max-width: 35rem`

simplq/src/components/common/QueueStats/index.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+  const creationTime = useMemo(() => {
+    if (!queueCreationTimestamp) return '';
+
+    const localTimeStamp = moment(queueCreationTimestamp); // TODO: Make sure the local time is always displayed
+    return `${localTimeStamp.format('LT')} ${localTimeStamp.format('ll')}`;
+  }, [queueCreationTimestamp]);
+
+  return (
+    <div>
```
daltonfury42, line 30: Can you remove this redundant div?

simplq/src/components/pages/Join/index.jsx
```diff
@@ -45,6 +46,7 @@ export default (props) => {
     <div>
       <HeaderSection queueName={queueStatusResponse.queueName} history={props.history} />
       <div className={styles['main-content']}>
+        <QueueStats queueStatus={queueStatusResponse} />
```
daltonfury42, line 49: In this or later pr, can we do something like:
```
<div className={styles['main-content']}>
  <div className={styles['section']}>
        <p className={styles['message']}>Please enter your contact details to join this queue</p>
        <JoinQueueForm
          queueId={queueId}
          joinQueueHandler={joinQueueHandler}
          buttonText="Join Queue"
        />
  </div>
  <div className={styles['section']}>
     <QueueStats queueStatus={queueStatusResponse} />
     < svg here >
  </div>
</div>
```

It would then look roughly like in our initial design:

image

You can ignore the background color of the right side, or have it, upto you. On winder screens the two sections can come side by side, but on mobile, they can then flex-wrap so that the stats go below.

What do you think?

daltonfury42 (changes requested):
Nicely done. Thanks for the PR.

simplq/src/components/pages/Join/index.jsx
mradenovic: This was the Idea. I just extracted the existing code that can be reused. Should we also render different background color based on status (ACTIVE, PAUSED, DELETED)?

simplq/src/components/common/QueueStats/queueStats.module.scss
mradenovic: Done

simplq/src/components/common/QueueStats/queueStats.module.scss
mradenovic: `align-items: center;` needs to stay for now, since elements are not the same height.

`justify-content: space-between;` works well.

simplq/src/components/common/QueueStats/queueStats.module.scss
mradenovic: Done, will do for now.

simplq/src/components/pages/Join/index.jsx
mradenovic: `styles['section']` loks like a very good solution. Should be designed and reused in all pages since currently pages don't look good on big screen in different sizes.

Screenshot 2021-01-24 at 22 41 26

Screenshot 2021-01-24 at 22 42 16

simplq/src/components/pages/Join/index.jsx
```diff
@@ -46,7 +46,9 @@ export default (props) => {
     <div>
       <HeaderSection queueName={queueStatusResponse.queueName} history={props.history} />
       <div className={styles['main-content']}>
-        <QueueStats queueStatus={queueStatusResponse} />
+        <div style={{ display: 'flex', justifyContent: 'center' }}>
+          <QueueStats queueStatus={queueStatusResponse} />
```
maaverik, line 50: Could you create a class for this style and use it from the join SCSS file? We're separating styles from components everywhere else.

maaverik (changes requested):

simplq/src/components/pages/Join/index.jsx
maaverik: Different colors based on status is not a bad idea. But I'm worried it would mess with the color scheme of the site. We're using this tint of blue everywhere and having a single page which is predominantly a different color seems out of place. If you have some ideas, could you post pictures of the design, so we can deliberate?

daltonfury42: Agreed. Let's not change the background colour entirely, but it will we good UX to highlight attention to PAUSED and DELETED as @maaverik said.

daltonfury42: > styles['section'] loks like a very good solution. Should be designed and reused in all pages since currently pages don't look good on big screen in different sizes.

This is a great idea, we should try to do this. If you can raise an issue regarding this, we can come up with a proposal and take this up.

simplq/src/components/pages/Join/index.jsx
mradenovic: That was an unfortunate commit 😊

simplq/src/components/pages/Join/index.jsx
mradenovic: I would leave it at this state for now. I will open an issue with some suggestions and we can go from there.

daltonfury42 (approved):

mradenovic:
Why is linter failing? Is that something on my end to setup?

daltonfury42:
> Why is linter failing? Is that something on my end to setup?

In actions on PRs from forks, the tokens have only read only access, and hence they fail. I just checked, and there seems to be a recent workaround for this: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

Till now, what we do is to merge PRs are manually run linter and fix issues manually.

maaverik (approved):

PR #483, opened 2021-01-24
Move to Auth0 for authentication

daltonfury42:
Resolved https://github.com/SimplQ/simplQ-frontend/issues/461

@maaverik I was trying out [auth0](https://auth0.com/) as an authentication/identity provider. It went very well, and I am happy with the setup.

Followed their [React Quickstart](https://auth0.com/docs/quickstart/spa/react) for the basic setup. I had to refactor our API calling logic a bit:

1. Defined a new hook called `useRequest` which returns a function that can be used from within components to make authenticated API calls.
2. All the request generation code is defined in `api/requestFactory` folder.

After the refactor, our code is more cleaner, to me. I've added comments with sample usage in the `useRequest.js` file.

TODO:

1. Backend side changes to verify the access token with auth0. Without this, backend calls after login will result in auth failure.
2. Home page customisations for logged in user and list of queues in new logic.


simplq/src/api/requestFactory/index.js
```diff
@@ -0,0 +1,4 @@
+import * as QueueRequestFactory from './queue';
+import * as TokenRequestFactory from './token';
+
+export { QueueRequestFactory, TokenRequestFactory };
```
maaverik, line 4: Could you change the name from QueueRequestFactory to something else, maybe just QueueRequests? Since it's not a factory, the names just feels out of place and I think it'll confuse people looking at the code.

simplq/src/api/useRequest.js
```diff
@@ -0,0 +1,48 @@
...
+// const MyAwesomeComponent = () => {
+//   const { requestMaker } = useRequest();
+//   ...
+//   requestMaker(QueueRequestFactory.create(textFieldValue))
+//     .then((response) => {
+//       // Handle response here
+//       ...
+//     });
+// }
+
```
maaverik, line 20: Kudos on the description

maaverik (commented):
I'm still on the fence about this refactor. The old format felt more intuitive to me, we call a service, and it handles making the appropriate request, it felt like better abstraction. Now, we have two imports in every component for requests and each component has to explicitly make the request. I would have preferred having the useRequest be used in the old service scripts we had.

simplq/src/components/pages/Home/MyQueues.jsx
```diff
@@ -4,15 +4,18 @@ import IconButton from '@material-ui/core/IconButton';
...
   const history = useHistory();
   const myQueues = useSelector((state) => state.appReducer.myQueues);
+  const { requestMaker } = useRequest();
+
   const handleDelete = (e, queue) => {
     // Don't trigger parent's onClick
     e.stopPropagation();
-    deleteQueue(queue.queueId).then(() => history.push('/'));
+    requestMaker(QueueRequestFactory.deleteQueue(queue.queueId)).then(() => history.push('/'));
   };
```
maaverik, line 19: deleteQueue used to handle modifying the global store too. Is that being done in this version?
`await getMyQueues().then((subResponse) => store.dispatch(setMyQueues(subResponse.queues)));`

daltonfury42:
Right, the old design was simpler, but we have to do this as Auth0 lets you get access token for making API calls only via a React Hook the it provides. So we would have to do this inside _**each component**_ as hooks can't be called outside of other hooks and functional components:

```
useEffect(() => {
  const getUserMetadata = async () => {
    const domain = "YOUR_DOMAIN";

    try {
      const accessToken = await getAccessTokenSilently({
        audience: `https://${domain}/api/v2/`,
        scope: "read:current_user",
      });

      const userDetailsByIdUrl = `https://${domain}/api/v2/users/${user.sub}`;

      const metadataResponse = await fetch(userDetailsByIdUrl, {
        headers: {
          Authorization: `Bearer ${accessToken}`,
        },
      });

      const { user_metadata } = await metadataResponse.json();

      setUserMetadata(user_metadata);
    } catch (e) {
      console.log(e.message);
    }
  };

  getUserMetadata();
}, []);
``` 
(Example from their [quick start](https://auth0.com/docs/quickstart/spa/react/02-calling-an-api))

That is super verbose, and the react way of solving this is to define your own custom hook. Auth0 haa given a sample [here](https://github.com/auth0/auth0-react/blob/master/EXAMPLES.md#4-create-a-useapi-hook-for-accessing-protected-apis-with-an-access-token). This looked horrible to me at first, but after asking around, hooks are the react way of solving this ([Ref](https://reactjs.org/docs/hooks-intro.html#motivation)). Hooks in react allows you to keep a common logic in functional components in one place 

Now theory aside, this design might not look very intuitive at first, but there are some advantages. 

There are two parts to our API handling logic now, 1) a `useRequest` hook and 2) `RequestFactory`.

The `useRequest` hook is write once and forget. It makes network requests from inside react components. The `RequestFactory` is responsible for generating different types of requests.

It's a bit of single responsibility principle in action, also it's easy to extend. When a new API is added, all you need to do is define it's request structure in the request factory, and it's ready to use:

```
requestFactory/queue.js
export const create = (queueName) => ({ method: 'post', url: '/queue', data: { queueName } });
```

To someone who is reading our code, we are hiding away all that network request making code behind the hook, and he would not need to understand this. All in all, if this design stays in the long term, the first PR might look a bit less intuitive / complex, but it will ease things from then on.

simplq/src/api/requestFactory/index.js
daltonfury42: It's a request factory. It's creating request objects. An example:

```
export const create = (queueName) => ({ method: 'post', url: '/queue', data: { queueName } });
```

It takes a queue name and creates a queue creation request object.

simplq/src/api/useRequest.js
daltonfury42: I feel a bit more commenting is required in the request factory side. Let me add

simplq/src/components/pages/Home/MyQueues.jsx
daltonfury42: This was breaking our pattern, and coupling stuff. That was a bad design choice from my side. I am undoing it in next PR.

In the component where we require this data, now we can just call getMyQueues when displaying queues to the user. (Straigtforward like any other API call, no redux, no setting it here an there)

daltonfury42:
TLDR; If we don't do this, our components will get cluttered with access token fetching code.

This is true even if we use Okta: https://github.com/okta/okta-oidc-js/issues/825

Let me know if you have any other solution in mind, though.

maaverik:
I see, so since Auth0 recommends using its services through hooks, we can't get the token outside components. Then, you're right, this is probably the best approach.

maaverik (approved):

Issue #476, opened 2021-01-17
Unauthorised admin page accesses should not result in page load


maaverik:
**Describe the bug**
If we try opening the admin page of a queue created by another signed-in user, the page opens, although backend info doesn't come through. We shouldn't even show the admin page in cases like this.

**To Reproduce**
Open the admin link of a queue created by a different signed-in user. The page opens up with options even though the member list doesn't load.

**Expected behavior**
It should clearly say "unauthorized access, please login to continue" or "you're not the owner of this queue".

daltonfury42:
One option is to have a "Unauthorised" page in your [routes](https://github.com/SimplQ/simplQ-frontend/blob/master/simplq/src/components/Layout/Routes.jsx) like [PageNotFound](https://github.com/SimplQ/simplQ-frontend/tree/master/simplq/src/components/pages/PageNotFound).

Then whenever we make a network call, if backend returns (401, Unauthorised), we should take the user to the Unauthorised page [here](https://github.com/SimplQ/simplQ-frontend/blob/master/simplq/src/services/axios.js).

mathkruger:
@maaverik Can I take this one?

tusharrajpoot:
@maaverik can i do this?

maaverik:
This project is not being actively maintained anymore. Thanks for offering to help!

tushars-code:
Can i do this?

Issue #422, opened 2020-12-07
Add an option to create a new queue/ entering new code if no queue by that code exists


HJain13:
**Is your feature request related to a problem? Please describe.**
I was just trying it out, and instead of clicking create button, clicked join button; 404 came up

**Describe the solution you'd like**
While 404 is serviceable, A page explicitly mentioning a queue doesn't exist and prompting to either create one or enter a new code might be a better UX

maaverik:
Adding a prompt to create a queue wouldn't be the best idea since people who want to join a queue wouldn't want to create one. In case they clicked mistakenly, they can go to the home page. But a page specifically saying that the entered queue doesn't exist and having an option to enter another queue name to join would be good. 

Or, more easily, we could modify the navigation to just show the popup saying the entered queue doesn't exist instead of trying to go to the join page. The same thing that happens when someone tries to create a queue that already exists. This way, users wouldn't have to keep going back and forth between pages in case of mistakes. 

@daltonfury42 what do you think?

daltonfury42:
This is what happened: @HJain13 wanted to try out SimplQ, so he entered a new queue name, and unconsciously clicked on the Join queue button instead of create queue. We served him 404 page, and since he made this mistake unconsciously, he is now wondering if the site is down.

![image](https://user-images.githubusercontent.com/11256376/101497233-c2ba5f80-3990-11eb-864a-edf76d96ab73.png)

Better UX is to handle this situation by, rather than saying "404, ERROR", tell him that the queue he tried to join does not exist, and ask if he wants to create a queue with the entered name instead. This is slightly better UX. Even if user unconsciously made a mistake, we should handle it smoothly, when possible.

daltonfury42:
Rough idea, instead serve this in such scenarios:

![image](https://user-images.githubusercontent.com/11256376/101498686-67896c80-3992-11eb-992c-fbd5a254dc4e.png)

AleemRehman:
@daltonfury42 I can pick this up?

daltonfury42:
Sure. What solution do you have in mind? Simplest way that I could think of is to check from the 404 page component if the url matches the regex `/j/[A-Za-z0-9-]*$` if yes, render different text.

It might be a good idea to build the regex in one place inside [this](https://github.com/SimplQ/simplQ-frontend/blob/master/simplq/src/components/common/utilFns.js#L7) file and reuse. What do you think?

AleemRehman:
[...]

PR #484, opened 2021-01-24
Creating unauthorized page

mathkruger:
Fix for issue #476

### DIFF
--- a/simplq/src/components/pages/Unauthorized/unauthorized.module.scss
+++ b/simplq/src/components/pages/Unauthorized/unauthorized.module.scss
@@ -1,0 +1,26 @@
+@import '../../common.scss';
+
+.main {
+  @include center-horizontally();
+  @include center-vertically();
+  @include text-color();
+  font-size: x-large;
+  min-height: 75vh;
+  flex: 1;
+  flex-direction: row;
+  flex-wrap: wrap;
+  div {
+    padding: 1rem;
+  }
+  Button {
+    cursor: pointer;
+  }
+  .errtext {
+    text-align: left;
+    padding: 0 40px;
+  }
+  .buttondiv {
+    display: flex !important;
+    justify-content: center;
+  }
+}
--- a/simplq/src/components/pages/Unauthorized/unauthorized.module.scss
+++ b/simplq/src/components/pages/Unauthorized/unauthorized.module.scss
@@ -20,7 +20,7 @@
     padding: 0 40px;
   }
   .buttondiv {
-    display: flex !important;
+    display: flex;
     justify-content: center;
   }
 }
--- a/simplq/src/components/pages/Unauthorized/unauthorized.module.scss
+++ b/simplq/src/components/pages/Unauthorized/unauthorized.module.scss
@@ -10,7 +10,7 @@
   flex-direction: row;
   flex-wrap: wrap;
   div {
-    padding: 1rem;
+    padding: .5rem;
   }
   Button {
     cursor: pointer;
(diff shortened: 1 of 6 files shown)


simplq/src/services/axios.js
```diff
@@ -14,6 +14,8 @@ const handleApiErrors = (err) => {
         `There's a problem with the data you've entered ${err.response.data.message}`
       )
     );
+  } else if (err.response.status === 401) {
+    window.location.href = `${window.location.origin}/unauthorized`;
   } else {
```
maaverik, line 19: @daltonfury42 what do you think about this approach? Should we go ahead with this or conditionally switch to the unauthorized page from within the admin component?

maaverik (changes requested):

maaverik:
Another thing is, this page is almost identical to the 404 page we have. Maybe we could make a standard error page component and pass in props for text, buttons and image for both the 404 page and this unauthorized page.

maaverik:
Speaking of images, we could add an image for this page too. I got this image from `https://undraw.co/` titled _secure login_.

Screenshot from 2021-01-25 15-26-59

Maybe you could add this image or another one that makes sense here.

simplq/src/services/axios.js
daltonfury42: I actually don't know. I am trying to think of what could go wrong in our current implementation if we go this route. 🤔 

If we want to conditionally switch to unauthorised page from admin component, we need to let admin component know that it was 401. But we took the call to handle API errors at one place in here, so it might be difficult...(It's a design limitation). Do you see a way?

If not, can we go ahead with this, and then see if it doesn't cause problems? (I can't see any right now, though it might give issues later)

daltonfury42:
> Another thing is, this page is almost identical to the 404 page we have. Maybe we could make a standard error page component and pass in props for text, buttons and image for both the 404 page and this unauthorized page.

If we are planing to add CreateJoinForm to the 404 page, it might come in the way, so we could keep them separate for now, and combine them later? @maaverik 

https://github.com/SimplQ/simplQ-frontend/issues/422

simplq/src/services/axios.js
maaverik: Okay let's do this for now and change later if needed.

simplq/src/components/pages/Unauthorized/unauthorized.module.scss
```diff
@@ -0,0 +1,26 @@
...
+  Button {
+    cursor: pointer;
+  }
+  .errtext {
+    text-align: left;
+    padding: 0 40px;
+  }
+  .buttondiv {
+    display: flex !important;
+    justify-content: center;
```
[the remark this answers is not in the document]
daltonfury42: Yes, let's remove. 👍

maaverik:
> > Another thing is, this page is almost identical to the 404 page we have. Maybe we could make a standard error page component and pass in props for text, buttons and image for both the 404 page and this unauthorized page.
> 
> If we are planing to add CreateJoinForm to the 404 page, it might come in the way, so we could keep them separate for now, and combine them later? @maaverik
> 
> #422

Yes, forgot about that. Let's keep them separate.

maaverik:
Could you merge with the current master to take care of these merge issues? Also, any thoughts on adding an image?

daltonfury42:
@mathkruger are you looking to complete the PR?

maaverik:
Closing this. @mathkruger you can reopen if you want to resume sometime.

PR #486, opened 2021-01-25
Modifed the 404 page to support queue creation if queue doen't exist

### LINKED ISSUE #422: Add an option to create a new queue/ entering new code if no queue by that code exists

maaverik:
Resolves #422 

I'm not sure if I'm missing something about the condition to reach this page. Currently, it shows the form if the join page errors out for any reason. I'm not sure if it's required, but we might have to add some extra conditions to check if the reason for the join page erroring out is because of the queue not existing.

Also, I'm open for suggestions on the text. The queue name is prefilled by the way.
Screenshot from 2021-01-25 17-08-18


daltonfury42:
Can you merge master branch to your branch?

simplq/src/components/common/CreateJoinForm/index.jsx
```diff
@@ -1,14 +1,15 @@
...
+import { handleEnterPress } from '../../../utils/eventHandling';
+import { isQueueNameValid } from '../../../utils/textOperations';
+import InputField from '../InputField';
+import LoadingIndicator from '../LoadingIndicator';
+import StandardButton from '../Button';
 
-const CreateJoinForm = () => {
-  const [textFieldValue, setTextFieldValue] = useState('');
+const CreateJoinForm = (props) => {
+  const [textFieldValue, setTextFieldValue] = useState(props.value);
```
daltonfury42, line 12: Can you rename `props.value` to `props.defaultTextFieldValue`?

simplq/src/components/common/CreateJoinForm/createJoinForm.module.scss
```diff
@@ -0,0 +1,23 @@
+@import '../../common.scss';
+
+.create-join-form {
+  margin: auto; // auto for centering
+  padding-top: 2rem;
+  max-width: 35rem;
+  position: relative;
+  // z-index set greater than that of svg below,
```
daltonfury42, line 8: `s/svg below/svg on home page`

simplq/src/components/pages/PageNotFound/index.jsx
```diff
@@ -1,12 +1,29 @@
...
 import Button from '../../common/Button';
 import styles from './pageNotFound.module.scss';
+import CreateJoinForm from '../../common/CreateJoinForm';
 
 function PageNotFound(props) {
-  return (
-    <>
+  const MainContent = () => {
+    const urlPath = window.location.href;
+    if (isJoinPageUrl(urlPath)) {
```
daltonfury42, line 10: In Join page, this is what we are doing:

```
  const [error, setError] = useState(false);
  useEffect(() => {
    async function fetchData() {
      const response = await QueueService.getStatusByName(queueName);
      if (response) {
      } else {
        setError(true);
      }
    }
    fetchData();
  }, [queueName]);

  if (error) {
    return <PageNotFound history={props.history} />;
  }
```

Here on error, we are rendering the 404 page (as if it was a component). Instead can we route to 404 page like this? I feel this is more proper:

```
  useEffect(() => {
    async function fetchData() {
      const response = await QueueService.getStatusByName(queueName);
      if (response) {
      } else {
        props.history.push('/queueNotFound/queueName='+queueName)
      }
    }
    fetchData();
  }, [queueName]);
```

Then define a new route which would also render the same 404 page:

```
<Route path="/pageNotFound/queueName=:queueName" exact component={QueueStatus} />
```
The catch all 404 route also stays.

Then you can conditionally check and retrieve the queueName from `props.match.params.queueName` here. Will this work?

daltonfury42:
> I'm not sure if I'm missing something about the condition to reach this page. Currently, it shows the form if the join page errors out for any reason. I'm not sure if it's required, but we might have to add some extra conditions to check if the reason for the join page erroring out is because of the queue not existing.

The unauthorized flow that @mathkruger is working on has the same issue. I feel our central error handling logic should show the error popup, and then return an object back to the caller [here](https://github.com/SimplQ/simplQ-frontend/blob/master/simplq/src/services/axios.js#L20) than an empty object, so that the caller can also take decisions based on the error.

simplq/src/components/pages/Home/home.module.scss
```diff
@@ -24,30 +24,7 @@
...
-      }
-    }
-    .input-box {
-      min-height: 5rem;
-    }
-  }
   svg {
-    // z-index set to lover value than upper create-join-form
-    z-index: 2;
     // global margin of 1 rem set. Setting negative margin to fill till the edge
```
maaverik, line 28: Looks like it works fine without z-index now, let me know if you notice any problems anywhere.

simplq/src/components/common/CreateJoinForm/index.jsx
```diff
@@ -1,23 +1,23 @@
...
+  const [textFieldValue, setTextFieldValue] = useState(props.defaultTextFieldValue);
   const [invalidMsg, setInvalidMsg] = useState('');
   const [createInProgress, setCreateInProgress] = useState(false);
   const { requestMaker } = useRequest();
   const history = useHistory();
 
   const handleCreateClick = () => {
-    if (textFieldValue === '') {
+    if (!textFieldValue) {
       setInvalidMsg('Queue name is required');
```
maaverik, line 21: The initial value comes in as undefined now, so made this change to prevent queues without names from being created.

maaverik (commented):
Let me know if you see any more problems

maaverik:
> > I'm not sure if I'm missing something about the condition to reach this page. Currently, it shows the form if the join page errors out for any reason. I'm not sure if it's required, but we might have to add some extra conditions to check if the reason for the join page erroring out is because of the queue not existing.
> 
> The unauthorized flow that @mathkruger is working on has the same issue. I feel our central error handling logic should show the error popup, and then return an object back to the caller [here](https://github.com/SimplQ/simplQ-frontend/blob/master/simplq/src/services/axios.js#L20) than an empty object, so that the caller can also take decisions based on the error.
> 
> Edit: The error object can have some fields like return code, errorText etc, and it should the thrown as a error so that it will be caught by catch statements downstream.

So basically modify the axios catch with something like this, right?
```
    .catch((e) => {
        handleApiErrors(e);
        return e;
    })
```
Yeah, I think it'll work fine. The error object has all the info we'd need.

simplq/src/components/pages/PageNotFound/index.jsx
```diff
@@ -1,12 +1,29 @@
...
 import Button from '../../common/Button';
 import styles from './pageNotFound.module.scss';
+import CreateJoinForm from '../../common/CreateJoinForm';
 
 function PageNotFound(props) {
-  return (
-    <>
+  const MainContent = () => {
+    const urlPath = window.location.href;
+    if (urlPath.includes('queueName=')) {
```
daltonfury42, line 9: Would this have worked?

```
if (props.match.params.queueName) {
  ...
}
```

daltonfury42 (approved):

simplq/src/components/pages/PageNotFound/index.jsx
maaverik: Yes, this is better

PR #491, opened 2021-01-26
Load site content only after auth exchange, remove min-height for loading indicator

daltonfury42:
See inline comments,

1. Load the main content only after auth is done. I went this route as otherwise we would have to check everywhere before making a call if login request is completed.
2. Removed min-height, @maaverik I could not find the place where we need a minimum height... let me generate a preview url, then can you tell where we need this height? I'll also look

### DIFF
--- a/simplq/src/components/Layout/index.jsx
+++ b/simplq/src/components/Layout/index.jsx
@@ -1,14 +1,16 @@
 import React from 'react';
+import { useAuth0 } from '@auth0/auth0-react';
 import Routes from './Routes';
 import Footer from '../common/Footer';
 import styles from './layout.module.scss';
+import LoadingIndicator from '../common/LoadingIndicator';
 
 function Layout() {
+  const { isLoading } = useAuth0();
+
   return (
     <div className={styles['box']}>
-      <div className={styles['content']}>
-        <Routes />
-      </div>
+      <div className={styles['content']}>{isLoading ? <LoadingIndicator /> : <Routes />}</div>
       <div className={styles['footer']}>
         <Footer />
       </div>
--- a/simplq/src/components/Layout/layout.module.scss
+++ b/simplq/src/components/Layout/layout.module.scss
@@ -8,6 +8,8 @@
 
   .content {
     flex: 1 1 auto;
+    display: flex;
+    flex-direction: column;
   }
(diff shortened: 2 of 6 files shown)


simplq/src/components/Layout/index.jsx
```diff
@@ -1,14 +1,16 @@
...
 
 function Layout() {
+  const { isLoading } = useAuth0();
+
   return (
     <div className={styles['box']}>
-      <div className={styles['content']}>
-        <Routes />
-      </div>
+      <div className={styles['content']}>{isLoading ? <LoadingIndicator /> : <Routes />}</div>
```
daltonfury42, line 13: I want to load the main site content only after auth is loaded. The entire page will have a loading indicator for a moment on page load, but this eases a lot of things

maaverik:
What I remember is the min-height caused problems with the join page. The spinner wasn't visible when the join page was loading. But, this preview doesn't seem to have that problem for some reason.

simplq/src/components/Layout/index.jsx
maaverik: I don't really understand the problem you've solved here, could you explain?

daltonfury42: Auth0 takes a small amount of time at page load to initialise. If we make a API call before this, it would result in auth failure from the backend API. There are other places, for example, we have to wait for it to load before deciding if we should render the login or logout button. Same with the temporary queue warning ribbon.

So it looked visually bad, with too much spinners and changes on page refresh. Also had to handle it in code everywhere.

So before we load the main page section, I thought of waiting till Auth0 "is loaded".

maaverik: Okay, got it. So you essentially made auth a sync call. That makes sense.

maaverik:
I guess you can merge this. We can figure out the problem with the spinner if it still exists the next time it occurs.

simplq/src/components/Layout/index.jsx
daltonfury42: Basically, I am waiting for auth to finish (at page load auth0 would get a fresh token from their server if user is already logged in, else it set itself to "not logged in") before rendering the content.

PR #493, opened 2021-01-26
Refactor component names

mradenovic:
- Rename `index.jsx` to `ComponentName.jsx`
- Use `PascalCase` for `ComponentName.module.scss`
- Use `index.jsx` to re-export component


mradenovic:
Only common components will be changed in this PR. Please comment before I do more work.

daltonfury42:
I did some research, and this makes sense.

https://stackoverflow.com/questions/44092341/how-do-index-js-files-work-in-react-component-directories

mradenovic:
The other option would be to get rid of all `index.jsx` files. Then imports would have to be updated too.

The only bad practice in my opinion is to have components within `index.jsx`. It should be what its name says, an index 😄

mradenovic:
common/Nav doesn't have `index.jsx`. A good example of how it works without it.

I was not sure if I should split `Nav.module.scss` in more files, since every component within the folder uses only its own css.

daltonfury42 (approved):

PR #506, opened 2021-01-30
Fix lint issues

### LINKED ISSUE #505: Enforce lint checks as premerge check
We had at some point removed the branch protection rule to allow merges only if lint checks passed. I think it was because too many new collaborators were finding it difficult to fix and merge, that we decided to disable it, and fix all at one go on out own.

Now we are at a stage were we can make one final fix and enable it back. 

https://github.com/SimplQ/simplQ-frontend/issues/503 would have been could pre-merge if this was in place.

daltonfury42:
Resolves #505


simplq/src/components/pages/Admin/ControlScroll.js
```diff
@@ -1,25 +1,37 @@
 // left: 37, up: 38, right: 39, down: 40,
 // spacebar: 32, pageup: 33, pagedown: 34, end: 35, home: 36
-let keys = {37: 1, 38: 1, 39: 1, 40: 1};
+const keys = { 37: 1, 38: 1, 39: 1, 40: 1 };
 
-const preventDefault = (e) => e.preventDefault();
+function preventDefault(e) {
+  e.preventDefault();
```
maaverik, line 6: Is there any specific reason why these were changed?

maaverik (approved):

simplq/src/components/pages/Admin/ControlScroll.js
daltonfury42: Arrow functions were giving lint issues. They should always return a value for eslint to be kept happy. That was cluttering the function with unwanted `return undefined;` which was looking confusing from readability point.

So then standardised to use `function` as other two functions in the module are already functions.

maaverik: Okay

Issue #477, opened 2021-01-17
Signed in users should be the owners of their existing queues created in the same session


maaverik:
**Is your feature request related to a problem? Please describe.**
If a user creates a temporary queue, and he signs in after that, the created queue is still not owned by anyone.

**Describe the solution you'd like**
All created ownerless queues in the same session as the user sign in should change to being owner by the user. This can be done by creating a random token when a temporary queue is created and send this token as the queue owner in the backend and store this token in SessionStorage. When the user signs in, all such tokens in SessionStorage are taken and the corresponding queues are changed to be owned by the current user.

navaneeth-spotnana:
As a first step, I've added a unique device UUID which will be sent with every unauthenticated request to the backend. So, from now own, every anonymous user will see only their own queues.

Now we need to add a backend endpoint to migrate all resources created against such a UUID to a user and call it after the user logs in.

PR #441, opened 2020-12-24
Add navbar to all Pages, nav minor bugfixes

daltonfury42:
I was trying adding navbar to all pages.

What triggered this was that we want the login button and all pages. It's easier to just make the navbar common for all pages.

Also some trivial bugfixes. Clicking on logo and "how it works" was not working, fixed those.
Removed some z-indexes along the way and used flexbox.

PR #507, opened 2021-01-31
Move logic to redux

### LINKED ISSUE #517: Refactor QueueDetails
Refactor `QueueDetails` component to use redux. Possibly rename to `QueueStatus` to match the server logic.

- [x] Create a new issue for the task and assign yourself or ask a maintainer to do so.
- [x] Rename (or better recreate) request-factory function to properly represent the action.
- [x] Make an async action matching the request.
- [x] Add nice jsDocs.
- [x] Create reducers for the needed state.
- [x] Redesign components to use redux state and actions.

### LINKED ISSUE #518: Refactor Join page component
Refactor `Join` page component to use redux.

- [x] Create a new issue for the task and assign yourself or ask a maintainer to do so.
- [x] Rename (or better recreate) request-factory function to properly represent the action.
- [x] Make an async action matching the request.
- [x] Add nice jsDocs.
- [x] Create reducers for the needed state.
- [x] Redesign components to use redux state and actions.

mradenovic:
Pros:
- [Strongly recomended](https://redux.js.org/style-guide/style-guide#put-as-much-logic-as-possible-in-reducers)
- Easier testing
- Easier mocking of libraries and use of storybook
- More flexible notification handling

Resolves #517
Resolves #518


daltonfury42:
Wonderful, I always wanted to move to using thunk and actions, but could never get it right. Redux toolkit way of doing this was a bit confusing to me. Let me study your PR.

mradenovic:
`redux-toolkit` was confusing for me too. Actually, it is very elegant. What makes things complicated is that `auth0' works only in components.

My idea is to pass `auth0` to any async thunk. To suppress `serializableCheck` warnings, it must be ignored. Convention to make things clear would be:

```js
const someAsyncAction = createAsyncThunk('action/requestStatus', async (arg) => {
  const { auth, payload } = arg;
  const { id } = payload
  const response = await doSomeethingById(auth, id);
  return response;
});

/**
 * Dispatching async actions
 *
 * When dispatching async actions with auth0,
 * wrap the `payload` within `arg` parameter of the `payloadCreator` function.
 *
 * Example:
 */

const auth = useAuth()
const payload = { id: '0' }
const arg = {
  auth: auth,
  payload: payload
}
dispatch(someAsyncAction(arg))
```

maaverik:
I've never worked with thunk, so I'll hold off on reviewing till I get a good understanding of it. But essentially, the idea here is to abstract away both the authentication and API calls to reducers, right @mradenovic ?

mradenovic:
> I've never worked with thunk, so I'll hold off on reviewing till I get a good understanding of it. But essentially, the idea here is to abstract away both the authentication and API calls to reducers, right @mradenovic ?

Yes. It is easier for maintenance, and if decided to change something in the logic (i.e. replace auth0 with [Firebase auth](https://firebase.google.com/products/auth) ), it can be done in one place.

simplq/src/store/queuesSlice.js
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable no-param-reassign */
+
+import { createAsyncThunk, createSlice } from '@reduxjs/toolkit';
+import { QueueRequestFactory } from 'api/requestFactory';
+import makeAuthedRequest from 'api/axios-alt';
+
+const fetchQueues = createAsyncThunk('queues/requestStatus', async (arg) => {
+  const { auth } = arg;
+  const authedRequest = makeAuthedRequest(auth, QueueRequestFactory.getMyQueues());
```
daltonfury42, line 9: We should centralise this into redux. Our components would then look so much more clean. Can you also add one more where there will be a payload?

Can you also make the change for the create queue button?

daltonfury42:
We should centralise this into redux. Our components would then look so much more clean. Can you also add one more where there will be a payload?

mradenovic:
> We should centralise this into redux. Our components would then look so much more clean. Can you also add one more where there will be a payload?

I am working on next itteration. I managed to completely move auth to async thunk with hook. It has some quirks though. I also discovered some bug. It looks like current delete functionality doesn't update myQueues. It worked before #504 was applied, but that is because that bug was constantly fetchin data from the server.

daltonfury42:
> I managed to completely move auth to async thunk with hook
Woah, how did you do that? I did find somewhere a custom middleware+local storage for token solution, but didn't like it.

Yes, I also noticed this. As a side effect of #504 it was getting updated before. But this issue would go away when we move to redux and have the delete queue action that would update the global store, I guess?

mradenovic:
One more curriosity here: The server does not check if user is authenticated. If I send a request from client as `anonyous`, I get all the queues (300+). Than I can delete any of the queues even if I am not the owner.

```js
// api/axios-alt.js

// to test, change line 16 declaration

  const accessToken = false //auth.isAuthenticated
    ? await auth.getAccessTokenSilently({ audience: 'https://devbackend.simplq.me/v1' })
    : 'anonymous';
```

daltonfury42:
Yes, this in in my list. I am thinking of doing this:

When a user is not logged in, we create a UUID and store it in the local storage. Instead of `Bearer anonymous`, we would send this ID in case of unauthenticated users. This ID would be used to identify the resources created by users who have not signed in.

Then post sign in, we call a backend API that would transfer ownerships from this ID to the signed in user, if such an ID exists in local storage.

What do you think? Ideally there should be some anonymous login, which would generate short lived tokens, but this setup looks good enough for me.

mradenovic:
> What do you think? Ideally there should be some anonymous login, which would generate short lived tokens, but this setup looks good enough for me.

I don't know much about handling users on the backend. I usually use ready made solutions when I need one 🤣 Firebase auth has all of that solved and that was what I used in the past for prototyping.

simplq/src/components/pages/Home/MyQueues.jsx
```diff
@@ -1,41 +1,39 @@
...
-import useRequest from '../../../api/useRequest';
 
 export default () => {
   const history = useHistory();
-  const { requestMaker } = useRequest();
-  const [myQueues, setMyQueues] = useState([]);
-  const { isAuthenticated } = useAuth0();
+  const dispatch = useDispatch();
+  const queues = useSelector(selectQueues);
+  const fetchQueues = useCallback(useFetchQueues(), []);
```
mradenovic, line 13: Use `useCallback()` only if the action is going to be dispatched in `useEffect()`.

simplq/src/components/pages/Home/MyQueues.jsx
```diff
@@ -1,41 +1,39 @@
...
-  }
+    dispatch(fetchQueues());
+  }, [dispatch, fetchQueues]);
 
   const handleDelete = (e, queue) => {
     // Don't trigger parent's onClick
+    const arg = { queueId: queue.queueId };
     e.stopPropagation();
-    requestMaker(QueueRequestFactory.deleteQueue(queue.queueId)).then(() => history.push('/'));
+    dispatch(deleteQueue(arg))
```
mradenovic, line 24: `arg` is the same as `payload`. From my understanding, while redux action can have a form `{ type: 'deleteQueue', id: queueId, name: queueName }`, [async thunk](https://redux-toolkit.js.org/api/createAsyncThunk#payloadcreator) can accept only one argument. This also can be called in this way obviously 😄 :

```js
// Folowing the wisdom that payload should always be a single value
const queueId = queue.queueId
dispatch(deleteQueue({ queueId }))

// or
dispatch(deleteQueue({ queueId: queue.queueId }))
```

daltonfury42: I'll review this tomorrow morning your time.

daltonfury42: yes, the change makes sense. I like to avoid references if it will be used just once:
```
dispatch(deleteQueue({ queueId: queue.queueId }))
```

simplq/src/components/pages/Home/MyQueues.jsx
```diff
@@ -1,41 +1,39 @@
...
+      // waht is the next line?
+      .then(() => history.push('/'));
```
daltonfury42, line 26: Once queue is deleted, we take the user back to home page. In the long term, we would show admin page for deleted queues as well, but we won't need it in near future.

simplq/src/components/pages/Status/QueueDetails.jsx
```diff
@@ -1,31 +1,29 @@
...
+  const getQueueStatus = useCallback(useGetQueueStatus(), []);
 
   useEffect(() => {
-    async function fetchData() {
-      const response = await requestMaker(QueueRequestFactory.getStatus(props.queueId));
-      setQueueStatusResponse(response);
-    }
-    fetchData();
-  }, [props.queueId, requestMaker]);
+    // TODO: Can this state be reused? Do we need a reducer for it?
```
daltonfury42, line 14: Queue status should also be stored in global state? What do you think?

simplq/src/store/appSlice.js
```diff
@@ -19,6 +20,26 @@ const appSlice = createSlice({
...
+    [deleteQueue.pending]: (state) => {
+      state.infoText = `Deleting queue...`;
+    },
+    [deleteQueue.rejected]: (state, action) => {
+      state.errorText = action.error.message;
+    },
+    [deleteQueue.fulfilled]: (state, action) => {
+      state.infoText = `Deleted ${action.payload.queueName}`;
+    },
+  },
```
daltonfury42, line 42: I'll understand and study this later. It should ideally render spinners at the right places. But can we can handle this later?

simplq/src/store/queues/queuesSlice.js
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable no-param-reassign */
+
+import { createSlice } from '@reduxjs/toolkit';
+import { fetchQueues, deleteQueue } from './queuesAsyncActions';
+
+const queuesSlice = createSlice({
+  name: 'queues',
+  initialState: [],
```
daltonfury42, line 8: There are three types of information that we would store for queues:

1) List of queues, this is what you have implemented here. (yeah, `UserQueues` looks like a better name)
2) Status of a single queue 
3) Detailed queue object (used in admin page)

Do we plan to have three different slices for these? What is the best approach?

simplq/src/store/queues/queuesAsyncActions.js
```diff
@@ -0,0 +1,101 @@
+/* eslint-disable no-param-reassign */
+import useAuth, { makeAuthedRequest } from 'api/auth';
+
+import { createAsyncThunk } from '@reduxjs/toolkit';
+import { QueueRequestFactory } from 'api/requestFactory';
+import * as RequestFactory from 'api/requestFactory';
```
daltonfury42, line 6: We will standardise here to one or the other?

daltonfury42 (commented):
Still understanding the change, thanks!

simplq/src/api/requestFactory/index.js
```diff
@@ -17,3 +17,4 @@ import * as QueueRequestFactory from './queue';
 import * as TokenRequestFactory from './token';
 
 export { QueueRequestFactory, TokenRequestFactory };
+export { getQueueStatus } from './queue';
```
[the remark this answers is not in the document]
mradenovic: I recognized the idea, just wanted to make it shorter and more uniform.

```js
import * as RequestFactory from 'api/requestFactory';

RequestFactory.getQueueStatus(queueId)
RequestFactory.getToken(tokenId)

// instead of 
QueueRequestFactory.getQueueStatus(queueId)
TokenRequestFactory.getToken(tokenId)
```

simplq/src/api/requestFactory/queue.js
```diff
@@ -11,6 +11,16 @@ export const get = (queueId) => ({ method: 'get', url: `/queue/${queueId}` });
 // This endpoint can be envoked by non-owners as well.
 export const getStatus = (queueId) => ({ method: 'get', url: `/queue/status?queueId=${queueId}` });
 
+/**
+ *
+ * @param {string} queueId
+ * @returns {Object} request - partial axious request without baseURL
```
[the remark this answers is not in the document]
mradenovic: >Can you add a comment to delete getStatus defined above?

I haven't check if it was used anywhere else, I guess if I delete it, I will now 🤣

simplq/src/components/pages/Home/MyQueues.jsx
mradenovic: The user is already on the home page. No need for this, will be removed.

simplq/src/components/pages/Status/QueueDetails.jsx
mradenovic: We can decide later. It makes sense only if it can be reused from the state. Maybe it can be preserved when going from the Join page to the Status page.

simplq/src/store/appSlice.js
mradenovic: That's the real beauty of this. We can:
- match all pending requests and have only one state with one or many spinners looking at it;
- have separate loading state and spinner for every async action;
- make one async action wait for another, that waits for another... and update state when it's all over;
- DO TI LATER 😄

simplq/src/store/queues/queuesAsyncActions.js
mradenovic: I am leaning towards one `RequestFactory`, as described [earlier](https://github.com/SimplQ/simplQ-frontend/pull/507/files#r569711806).

simplq/src/store/queues/queuesSlice.js
mradenovic: After a quick look at data:
```js
const state = {
  queues: [],
  selectedQueue: {}, // It can be Admin queue, Join page queue, or status page Queue
  queueStatus: {}, //Updated when selectedQueue is updated
  tokens: [] // Updated with Admin page queue
}
```

mradenovic: we could expand on any of the above slices. Rearrange, extract, normalize...

```js
const state = {
  queues: {
    loading: false,
    ids: [],
    byId: {}
  },
  selectedQueue: {},
  queueStatus: {}, 
  tokens: {
    ids: [],
    byId: {}
  }
}
```

daltonfury42: Interesting, why store `ids: []`? Isn't it already covered by `byId: {}`?

simplq/src/components/pages/Status/QueueDetails.jsx
daltonfury42: For me it's more about uniformity.

simplq/src/store/queues/queuesSlice.js
mradenovic: At this stage, it doesn't really matter. For more complex future [normalizing state](https://redux.js.org/recipes/structuring-reducers/normalizing-state-shape) and [connecting more components](https://redux.js.org/recipes/structuring-reducers/normalizing-state-shape) is recommeded.

simplq/src/components/pages/Status/QueueDetails.jsx
mradenovic: On second look, it needs to be in the store, and async actions should be split into separate files, possibly in its own folder.

simplq/src/store/queues/queuesAsyncActions.js
```diff
@@ -0,0 +1,101 @@
...
+ * @returns — fetchQueues async async action creator
+ *
+ * @example
+ *
+ * import useFetchQueues from 'store/queues'
+ *
+ * const fetchQueues = useFetchQueues()
+ */
+const useFetchQueues = () => {
+  const auth = useAuth();
```
daltonfury42, line 20: this is pretty smart ✌️

simplq/src/store/queues/queuesAsyncActions.js
```diff
@@ -0,0 +1,101 @@
...
+
+  const fetchQueues = createAsyncThunk('queues/requestStatus', async () => {
+    if (!auth || !auth.isAuthenticated) {
+      return { queues: [] };
+    }
```
daltonfury42, line 25: Do we need `!auth` here? We should also add `auth. isLoading` here, I think.

Also leave this comment:
```
// TODO: Remove `!auth.isAuthenticated` after resolving https://github.com/SimplQ/simplQ-frontend/issues/477 
```

simplq/src/store/queues/queuesAsyncActions.js
```diff
@@ -0,0 +1,101 @@
...
+};
+
+/**
+ * A shadow function that returns fetchQueues async action creator.
+ *
+ * This should be used just for the action creators that it generates.
+
+ * @see ./queuesSlice
+ *
+ * @see https://redux-toolkit.js.org/api/createAsyncThunk#return-value
```
daltonfury42, line 41: I'll have to read this 😄

Vishv07: @daltonfury42 where can I find deployed PR?

mradenovic: I'm sorry, I didn't understand the last one.

simplq/src/store/queues/queuesAsyncActions.js
mradenovic: We definitely do not need `!auth`, it is guaranteed to be there by the hook. There was a lot of copy-paste happening so there will be a lot to remove 😊 I haven't explored what exactly should happen when not authenticated, so that should be implemented later.

simplq/src/store/queues/queuesAsyncActions.js
daltonfury42: @mradenovic We generate temporary preview deployments. This can be triggered by putting any label on the PR (You can see a sample [here](https://github.com/SimplQ/simplQ-frontend/pull/441))

But due to some security limitations, this is not possible on PRs from forks.

mradenovic: OK, I'll push to repo to see how it works.

simplq/src/store/asyncActions/getUserQueues.js
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * import useFetchQueues from 'store/queues'
+ *
+ * const fetchQueues = useFetchQueues()
+ */
+const useGetUserQueues = () => {
+  const auth = useAuth();
+
+  const getUserQueues = createAsyncThunk(typePrefix, async () => {
+    if (!auth || !auth.isAuthenticated) {
```
daltonfury42, line 22: > We definitely do not need !auth, it is guaranteed to be there by the hook. There was a lot of copy-paste happening so there will be a lot to remove 😊 I haven't explored what exactly should happen when not authenticated, so that should be implemented later.

let's add a comment here for my reference:
```
// TODO: Remove `!auth.isAuthenticated` after resolving https://github.com/SimplQ/simplQ-frontend/issues/477 
```

also lets remove `!auth` and add `auth.isLoading` to the check

simplq/src/store/asyncActions/getQueueStatusByName.js
```diff
@@ -0,0 +1,41 @@
...
+const useGetQueueStatusByName = () => {
+  const auth = useAuth();
+  const history = useHistory();
+
+  const getQueueStatusByName = createAsyncThunk(
+    typePrefix,
+    async ({ queueName }, { rejectWithValue }) => {
+      if (!auth || !auth.isAuthenticated) {
+        return {};
+      }
```
daltonfury42, line 22: Here we should check just for `auth.isLoading`, as the queue status endpoint doesn't need authentication.

simplq/src/store/asyncActions/getQueueStatus.js
```diff
@@ -0,0 +1,30 @@
+import { createAsyncThunk } from '@reduxjs/toolkit';
+import useAuth, { makeAuthedRequest } from 'api/auth';
+import * as RequestFactory from 'api/requestFactory';
+
+const typePrefix = 'getQueueStatus/requestStatus';
+
+/**
+ * A hook to access the deleteQueue async action creator.
```
daltonfury42, line 8: it should be `getQueueStatus async action creator`.

Should we be naming this `getQueueStatusById`?

simplq/src/store/asyncActions/getQueueStatus.js
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @returns — getQueueStatus async action creator
+ */
+const useGetQueueStatus = () => {
+  const auth = useAuth();
+
+  const getQueueStatus = createAsyncThunk(typePrefix, async ({ queueId }) => {
+    if (!auth || !auth.isAuthenticated) {
+      return {};
+    }
```
daltonfury42, line 18: same as for getQueueStatusByName

simplq/src/store/asyncActions/getQueueStatusByName.js
```diff
@@ -0,0 +1,41 @@
...
+
+      const authedRequest = makeAuthedRequest(auth, RequestFactory.getQueueStatusByName(queueName));
+
+      try {
+        const response = await authedRequest;
+        return response;
+      } catch (error) {
+        history.push(`/pageNotFound/queueName=${queueName}`);
+        return rejectWithValue({ message: `Queue ${queueName} does not exist, try again...` });
```
daltonfury42, line 31: 👀

simplq/src/store/asyncActions/deleteQueue.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @returns — deleteQueue async async action creator
+ */
+const useDeleteQueue = () => {
+  const auth = useAuth();
+
+  const deleteQueue = createAsyncThunk(typePrefix, async (arg) => {
+    if (!auth || !auth.isAuthenticated) {
+      return { queues: [] };
+    }
```
daltonfury42, line 18: Here we should check just for auth.isLoading, this should work as in case the user has not logged in, it will send "Bearer anonymous" for now. Later I will handle it along with https://github.com/SimplQ/simplQ-frontend/issues/477

simplq/src/components/pages/Status/QueueDetails.jsx
```diff
@@ -1,31 +1,29 @@
...
     <SidePanelItem
       Icon={InfoIcon}
       title="Queue Details"
       description="Other information about the queue"
       expandable
-      loading={!queueStatusResponse}
+      loading={!queueStatus}
     >
-      <QueueStats queueStatus={queueStatusResponse} />
+      <QueueStats queueStatus={queueStatus} />
```
daltonfury42, line 26: This won't work, we would have to check for empty object. 🤔 

An isLoading flag might be better, as you were implementing before?

These kind of glitches we can fix in a follow up PR as well.

daltonfury42 (commented):
Can't wait to merge this. It will be massive.  💪 

For me weekend is here, I will migrate remaining components as soon as this PR goes in.

simplq/src/store/asyncActions/getUserQueues.js
mradenovic: Removing it everywhere. All checks of the status should be done in reducers triggered by `.pending``.fulfilled`, ` and `.rejected` actions.

simplq/src/store/asyncActions/getQueueStatus.js
mradenovic: In some other language, I would probably just overload the `getQueueStatus(queueID)`. Let's keep it as is for now.

simplq/src/components/pages/Status/QueueDetails.jsx
mradenovic: All `pending', `fulfilled`, and `rejected` should be handled in reducers. It can be done later.

mradenovic:
I suggest we continue the conversation at #522.

simplq/src/components/pages/Status/QueueDetails.jsx
Vishv07: code looks much cleaner now

PR #510, opened 2021-02-02
added share option

### LINKED ISSUE #509: Add a new button on the admin page to share queue info to different platforms
Right now, we have a "copy queue link" button on our admin page that copies text in the form `Hi! Use https://www.simplq.me/j/dfgbfhb to join my queue and get live updates.` to the clipboard. The user then shares this link manually on different apps or paltforms. 

Let's create a new button called `Share` that opens up a dialog showcasing different social media buttons and allowing the user to share the queue information directly to those platforms.  We could use packages like https://www.npmjs.com/package/react-share or https://www.npmjs.com/package/react-social-sharing or something similar. Ideally, we want to showcase some popular platforms like facebook, twitter etc. when users click it on their desktop while on mobile, all communication apps installed should show up.

There's a related discussion [here](https://github.com/SimplQ/simplQ-frontend/issues/480#issuecomment-770161075). Let's start by deciding how the dialog should look like and what buttons to show on desktop use. Then let's figure out how to display all installed communication apps to be shown to the user on mobile.

prabureddy:
Resolves #509

### DIFF
--- a/simplq/.eslintrc.json
+++ b/simplq/.eslintrc.json
@@ -19,6 +19,7 @@
     "react/destructuring-assignment": 0,
     "prefer-destructuring": 0,
     "import/no-named-as-default": 0,
-    "dot-notation": 0
+    "dot-notation": 0,
+    "react/jsx-props-no-spreading": 0
   }
 }
--- a/simplq/.eslintrc.json
+++ b/simplq/.eslintrc.json
@@ -19,7 +19,6 @@
     "react/destructuring-assignment": 0,
     "prefer-destructuring": 0,
     "import/no-named-as-default": 0,
-    "dot-notation": 0,
-    "react/jsx-props-no-spreading": 0
+    "dot-notation": 0
   }
 }
--- a/simplq/src/components/pages/Admin/ShareQueue.jsx
+++ b/simplq/src/components/pages/Admin/ShareQueue.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useRef } from 'react';
+import React, { useState, useRef, memo } from 'react';
 import { CopyToClipboard } from 'react-copy-to-clipboard';
 import FileCopyIcon from '@material-ui/icons/FileCopy';
 import { useDispatch } from 'react-redux';
@@ -25,8 +25,8 @@ import StandardButton from '../../common/Button';
 import { setInfoPopupMessage } from '../../../store/appSlice';
 import styles from './admin.module.scss';
 
-const CopyButton = (props) => {
-  const link = `${window.location.origin}/j/${props.queueName}`;
+const CopyButton = ({ queueName, tourTag }) => {
+  const link = `${window.location.origin}/j/${queueName}`;
   const dispatch = useDispatch();
 
   const handleShareButtonClick = () => {
@@ -35,18 +35,14 @@ const CopyButton = (props) => {
 
   return (
     <CopyToClipboard text={link}>
-      <StandardButton
-        onClick={handleShareButtonClick}
-        icon={<FileCopyIcon />}
-        tourTag={props.tourTag}
-      >
+      <StandardButton onClick={handleShareButtonClick} icon={<FileCopyIcon />} tourTag={tourTag}>
         Copy Queue Link
       </StandardButton>
     </CopyToClipboard>
   );
 };
 
-const ShareQueue = (props) => {
+const ShareQueue = ({ queueName, tourTag }) => {
   const [open, setOpen] = useState(false);
   const anchorRef = useRef(null);
 
@@ -61,20 +57,20 @@ const ShareQueue = (props) => {
     setOpen(false);
   };
 
-  const quote = `Hi! Use ${window.location.origin}/j/${props.queueName} to join my queue and get live updates.`;
-  const link = `${window.location.origin}/j/${props.queueName}`;
+  const link = `${window.location.origin}/j/${queueName}`;
+  const quote = `Hi! Use ${link} to join my queue and get live updates.`;
 
   return (
-    <>
+    <div className={styles['share']}>
       <ButtonGroup
         variant="contained"
-        className={styles['button-group']}
+        className={styles['button-background']}
         ref={anchorRef}
         aria-label="split button"
       >
-        <CopyButton {...props} />
+        <CopyButton {...{ queueName, tourTag }} />
         <Button
-          className={styles['button-group']}
+          className={styles['button-background']}
           color="primary"
           size="small"
           onClick={handleToggle}
@@ -83,7 +79,7 @@ const ShareQueue = (props) => {
         </Button>
       </ButtonGroup>
       <Popper
-        style={{ zIndex: 100, backgroundColor: '#fff' }}
+        className={styles['popper']}
         open={open}
         anchorEl={anchorRef.current}
         transition
@@ -102,20 +98,28 @@ const ShareQueue = (props) => {
                 <MenuList>
                   Share
                   <MenuItem>
-                    <FacebookShareButton url={link} quote={quote} style={{ display: 'flex' }}>
-                      <FacebookIcon size={24} round style={{ marginRight: 10 }} />
+                    <FacebookShareButton
+                      url={link}
+                      quote={quote}
+                      className={styles['share-button']}
+                    >
+                      <FacebookIcon size={24} round className={styles['share-icon']} />
                       Facebook
                     </FacebookShareButton>
                   </MenuItem>
                   <MenuItem>
-                    <TwitterShareButton url={link} title={quote} style={{ display: 'flex' }}>
-                      <TwitterIcon size={24} round style={{ marginRight: 10 }} />
+                    <TwitterShareButton url={link} title={quote} className={styles['share-button']}>
+                      <TwitterIcon size={24} round className={styles['share-icon']} />
                       Twitter
                     </TwitterShareButton>
                   </MenuItem>
                   <MenuItem>
-                    <WhatsappShareButton url={link} title={quote} style={{ display: 'flex' }}>
-                      <WhatsappIcon size={24} round style={{ marginRight: 10 }} />
+                    <WhatsappShareButton
+                      url={link}
+                      title={quote}
+                      className={styles['share-button']}
+                    >
+                      <WhatsappIcon size={24} round className={styles['share-icon']} />
                       Whatsapp
                     </WhatsappShareButton>
                   </MenuItem>
@@ -125,8 +129,8 @@ const ShareQueue = (props) => {
           </Grow>
         )}
       </Popper>
-    </>
+    </div>
   );
 };
 
-export default ShareQueue;
+export default memo(ShareQueue);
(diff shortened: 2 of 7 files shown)


simplq/src/components/pages/Admin/ShareQueue.jsx
```diff
@@ -26,4 +46,87 @@ const ShareQueue = (props) => {
...
+          <Grow
+            {...TransitionProps}
+            style={{
+              transformOrigin: placement === 'bottom' ? 'center top' : 'center bottom',
+            }}
+          >
+            <Paper>
+              <ClickAwayListener onClickAway={handleClose}>
+                <MenuList>
+                  Share
```
daltonfury42, line 103: Let's remove this text, it's pretty obvious even without it. What do you think?

simplq/src/components/pages/Admin/ShareQueue.jsx
```diff
@@ -26,4 +46,87 @@ const ShareQueue = (props) => {
...
+                      <TwitterIcon size={24} round style={{ marginRight: 10 }} />
+                      Twitter
+                    </TwitterShareButton>
+                  </MenuItem>
+                  <MenuItem>
+                    <WhatsappShareButton url={link} title={quote} style={{ display: 'flex' }}>
+                      <WhatsappIcon size={24} round style={{ marginRight: 10 }} />
+                      Whatsapp
+                    </WhatsappShareButton>
+                  </MenuItem>
```
daltonfury42, line 121: Can we somehow move the styles into CSS? Can you check?

simplq/src/components/pages/Admin/ShareQueue.jsx
```diff
@@ -26,4 +46,87 @@ const ShareQueue = (props) => {
   );
 };
 
+const ShareQueue = (props) => {
+  const [open, setOpen] = useState(false);
```
daltonfury42, line 50: The component gets re-rendered in the background when we make an API call to the server in the background. (See [here](https://github.com/SimplQ/simplQ-frontend/blob/master/simplq/src/components/pages/Admin/index.jsx#L38))

This causes the popper to close. You can observe this by opening it, and waiting for less than 10s. If it's trivial to get around this, you can try, otherwise leave it as is. This can be handled later more elegantly after https://github.com/SimplQ/simplQ-frontend/pull/507.

simplq/src/components/pages/Admin/ShareQueue.jsx
```diff
@@ -26,4 +46,87 @@ const ShareQueue = (props) => {
...
+
+  const handleClose = (event) => {
+    if (anchorRef.current && anchorRef.current.contains(event.target)) {
+      return;
+    }
+    setOpen(false);
+  };
+
+  const quote = `Hi! Use ${window.location.origin}/j/${props.queueName} to join my queue and get live updates.`;
+  const link = `${window.location.origin}/j/${props.queueName}`;
```
daltonfury42, line 65: nit: Define link first and use it inside quote.

daltonfury42 (commented):
Neatly done. Thanks.

simplq/src/components/pages/Admin/ShareQueue.jsx
prabureddy: We can keep it

simplq/src/components/pages/Admin/ShareQueue.jsx
prabureddy: yes

simplq/src/components/pages/Admin/ShareQueue.jsx
prabureddy: done

simplq/src/components/pages/Admin/ShareQueue.jsx
prabureddy: Ok solved

simplq/src/components/pages/Admin/ShareQueue.jsx
```diff
@@ -1,29 +1,136 @@
...
+  WhatsappShareButton,
+  WhatsappIcon,
+} from 'react-share';
 import StandardButton from '../../common/Button';
 import { setInfoPopupMessage } from '../../../store/appSlice';
+import styles from './admin.module.scss';
 
-const ShareQueue = (props) => {
-  const shareUrl = `Hi! Use ${window.location.origin}/j/${props.queueName} to join my queue and get live updates.`;
+const CopyButton = ({ queueName, tourTag }) => {
```
daltonfury42, line 28: Lets pass `link` as a whole as props

simplq/src/components/pages/Admin/ShareQueue.jsx
daltonfury42: Alright, you mean that it is not helping, right? If not, let's remove the `memo()`

simplq/src/components/pages/Admin/admin.module.scss
```diff
@@ -8,6 +8,26 @@
...
+    z-index: 100;
+    background-color: #fff;
+  }
+
+  .share-button {
+    display: flex;
+  }
+
+  .share-icon {
+    margin-right: 10px;
```
daltonfury42, line 27: can you change to rem here. @maaverik likes rems. 😉

prabureddy: Ok sure :grinning:

prabureddy:
Updated!!!! 💯

simplq/src/components/pages/Admin/admin.module.scss
```diff
@@ -8,6 +8,26 @@
   height: max-content;
 }
 
+.share {
+  .button-background {
+    background-color: $primary-color-dark;
+    border-radius: 0.6rem !important;
+  }
```
maaverik, line 15: Is this !important required? I didn't notice any problems without it.

simplq/src/components/pages/Admin/admin.module.scss
```diff
@@ -8,6 +8,26 @@
...
+
+  .popper {
+    z-index: 100;
+    background-color: #fff;
```
maaverik, line 19: This z-index as well, I didn't see any problems without it

simplq/src/components/pages/Admin/admin.module.scss
```diff
@@ -8,6 +8,26 @@
...
+  }
+
+  .share-button {
+    display: flex;
+  }
```
maaverik, line 24: Same here

maaverik (commented):
This is a great start, kudos @prabureddy. We can brainstorm and improve it further down the line. A small addition I thought would be nice is if we could add a small border or a gradient difference between the dropdown arrow and the "Copy Queue Link" text, to make it really obvious that the arrow does something else, like we see in the fork and star button vs count in github.
Screenshot from 2021-02-03 00-48-37

simplq/src/components/pages/Admin/ShareQueue.jsx
maaverik: I noticed this too, let me see if I can figure out why.

maaverik: It's due to us polling for updates, some logic like this would work, atleast it looks like there are no problems when I tried it.
```suggestion
  let prevData;
  const update = useCallback(() => {
    clearTimeout(timeoutId);
    requestMaker(QueueRequestFactory.get(queueId)).then((data) => {
      if (data && JSON.stringify(prevData) !== JSON.stringify(data)) {
        setTokens(data.tokens);
        ...
```

Basically comparing old data with new one and only changing state if needed. There might be some inbuilt method that react provides for this, not really sure yet.

daltonfury42: A positive side effect of #507 is that, then one component will fire an action (refresh button, periodic refresh service), and only the component that subscribes to this data gets re-rendered. 

Then we would not have this problem. If we can wait till #507 is done, we can leave it as is and merge this PR. I don't see an urgent need to update main site. Else if we are going to do some fix, let us leave a comment here explaining what and why, so that we remove it post #507 .

mradenovic: #507 has its first component ready. To do all the components it will take more time or hands on deck. It should be a priority since it's gonna make a lot of things easier.

daltonfury42: @mradenovic Can you do one component and let us finalise and merge that. Once that's done, me, @maaverik and others will migrate the remaining components taking your PR as a reference.

I think we can do a lot in the same way. You lead the way and do one sample. Remaining repetitive work, we will take up.

Vishv07: Absolutely @daltonfury42  , Anything for me?

simplq/src/components/pages/Admin/admin.module.scss
prabureddy: I will update you now. Let complete and close this PR today

prabureddy: Yes !important is not required now. When I started this issue I noticed some other border-radius CSS already declared so included !important to overide it. After I have changed some config !important is not required but I forgot to remove it.

simplq/src/components/pages/Admin/admin.module.scss
prabureddy: https://github.com/SimplQ/simplQ-frontend/blob/9c0d900cfa501fae6c4aee61f0831cbffa327f93/simplq/src/components/common/SidePanel/SidePanel.module.scss#L1-L8

As here we mentioned z-index: 10 for the side panel how when the popper opens it hides behind the side panel so I included z-index: 100 in the popper

Screenshot from 2021-02-03 18-35-42

As you can see in this image the popper hides behind side panel

simplq/src/components/pages/Admin/admin.module.scss
prabureddy: Without  display: flex;
Screenshot from 2021-02-03 18-49-29

with display: flex;
Screenshot from 2021-02-03 18-49-40

with display: flex gives a good alignment

simplq/.eslintrc.json
```diff
@@ -19,6 +19,7 @@
     "react/destructuring-assignment": 0,
     "prefer-destructuring": 0,
     "import/no-named-as-default": 0,
-    "dot-notation": 0
+    "dot-notation": 0,
+    "react/jsx-props-no-spreading": 0
```
daltonfury42, line 23: Can we avoid this and explicitly pass down props?

I was [checking](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-props-no-spreading.md), and props spreading does affect readability.

If there is a good reason you need to do this in the PR, we can disable it inline where you are using it, but try to avoid.

simplq/src/components/pages/Admin/ShareQueue.jsx
daltonfury42: Yes, can you review #507?

simplq/src/components/pages/Admin/admin.module.scss
prabureddy: Yes removed both the z-index and everything looks fine

simplq/.eslintrc.json
prabureddy: Yes, removed (react/jsx-props-no-spreading) and passed down props. Now it is readable.

daltonfury42 (approved):

maaverik (approved):
Great PR. Thanks!

PR #515, opened 2021-02-03
Dynamically declare BASE_URL

chalx:
If `BASE_URL` environment variable exists it will be used, else fallback to the one from .env file


simplq/.env
```diff
@@ -0,0 +1 @@
+BASE_URL="http://localhost:8080"
```
daltonfury42, line 1: Lets fallback to `https://devbackend.simplq.me/v1`

simplq/package.json
```diff
@@ -36,13 +36,14 @@
     "styled-components": "^5.2.1"
   },
   "scripts": {
-    "start": "react-scripts start",
-    "build": "react-scripts build",
+    "start": "npm run config && react-scripts start",
+    "build": "npm run config && react-scripts build",
```
daltonfury42, line 40: We are configuring the url before build. Can we configure UI to take up the backend url dynamically after build?

Easiest way I could think of is to generate a json config object in the public folder and vary it from inside the frontend server. Then the client would request for this json object and derive the backend URL at start. 

I don't know if will lead to issues where a backend request happens before the backend URL is resolved.

chalx: @daltonfury42 I don't think it should be so complicated.

daltonfury42 (approved):
Let's merge this, and optimise later. Thanks

PR #521, opened 2021-02-04
Use node streams to generate config files

chalx:


chalx:
@daltonfury42 I have rewrite the generation process of config file

daltonfury42 (approved):
Yes, this looks nice.

PR #526, opened 2021-02-07
Use async actions for token status page

daltonfury42:
Move Token Status Page to use async actions and load state from redux.

Todo:
- Periodic refresh logic
- Minor refactors


simplq/src/store/asyncActions/getToken.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * A hook to access the getToken async action creator.
+ *
+ * @returns — getToken async action creator
+ */
+const useGetToken = () => {
+  const auth = useAuth();
+
+  const getToken = useCallback(
```
daltonfury42, line 16: @mradenovic Shall we move the useCallback into the action creator hook? Or is there a good reason to use it when called?

daltonfury42:
This is how I could think of doing periodic refresh. Please let me know if you see a better way to do this:

https://github.com/SimplQ/simplQ-frontend/pull/526/commits/ae2b981c37f7924ac453af4854d3e95fde74342d

simplq/src/store/asyncActions/getToken.js
mradenovic: `useCallback` should be used only if an action is dispatched from the `useEffect`. Look at the Join page for an example.

simplq/src/components/pages/Status/index.jsx
```diff
@@ -1,76 +1,47 @@
...
-        }
-      }
-      timeoutId = setTimeout(update, TIMEOUT);
-    });
-    // eslint-disable-next-line
-  }, [tokenId, oldTokenStatus]); // don't add showNotification, will result in infinite loop
+  const dispatch = useDispatch();
+  const token = useSelector(selectToken);
+  // const token = {}
+  const getToken = useGetToken();
```
mradenovic, line 18: `useGetToken()` should probably be wrapped in `useCallback`.

simplq/src/components/pages/Status/index.jsx
```diff
@@ -1,76 +1,47 @@
...
+  const deleteToken = useGetToken();
 
   useEffect(() => {
-    update();
-    return () => clearTimeout(timeoutId);
-  }, [update]);
+    dispatch(getToken({ tokenId, refresh: true }));
+  }, [tokenId, dispatch, getToken]);
```
mradenovic, line 23: `getToken()` her might cause fetch request on every render. Not 100% sure, please check, network tab, or just put some `conosle.log()` statement to see what's going on.

simplq/src/store/queueStatus/queueStatusSlice.js
```diff
@@ -13,8 +13,8 @@ const queueStatusSlice = createSlice({
     [getQueueStatusByName.fulfilled]: (state, action) => {
       return action.payload;
     },
-    // handle peding request
-    [getQueueStatus.pnding]: () => {
+    // handle pending request
+    [getQueueStatus.pending]: () => {
```
mradenovic, line 17: Good catch!

simplq/src/store/asyncActions/getToken.js
```diff
@@ -0,0 +1,40 @@
...
+          timer = setTimeout(() => dispatch(getToken({ tokenId, refresh })), REFRESH_INTERVAL);
+        }
+        return resp;
+      });
+      return response;
+    }),
+    []
+  );
+
+  return getToken;
```
mradenovic, line 35: If we decide to use `useCallback()` in action, it would be better to do it in `return` statement to avoid to many nested functions.

```js
return useCallback(getToken);
```

simplq/src/components/pages/Status/index.jsx
daltonfury42: Why do you think we should do here? I thought it's better to do it once inside the action than do it everytime the action than every time it is used?

simplq/src/store/asyncActions/getToken.js
daltonfury42: Is there a problem if we use it everytime?

mradenovic: To be honest, I am not sure, I tried to research with no final conclusion. I think every `useCallback` call will grab some memory. And I don't want to pollute async actions with an unnecessary wrapper.

simplq/src/components/pages/Status/index.jsx
mradenovic: Let's merge this as is. It will make maintenance easier, and if we can polish it later.

daltonfury42: I've standardised for now: https://github.com/SimplQ/simplQ-frontend/pull/526/commits/bb8f29ab1564f47e8995196071731083a5c2c331

The PR will be merged on your approval.

maaverik (approved):

PR #527, opened 2021-02-07
Refactor Token Status page to Take Data from Store Where required

daltonfury42:
Refactor to fetch from store where data is needed, than to pass it down from main component.

- Added a `loaded` boolean field into state. Thinking of setting just this flag while loading from a network call, than setting the whole state to `{}`.
- We were loading data in the root component, wanted to see how it would look like if we fetch from store only where data is required.

Looking for feedback

### DIFF
--- a/simplq/src/components/pages/Status/LeaveQueue.jsx
+++ b/simplq/src/components/pages/Status/LeaveQueue.jsx
@@ -2,22 +2,23 @@ import React from 'react';
 import { useDispatch } from 'react-redux';
 import CloseIcon from '@material-ui/icons/Close';
 import { useHistory } from 'react-router-dom';
+import { useGetToken } from 'store/asyncActions';
 import SidePanelItem from '../../common/SidePanel/SidePanelItem';
 import { setInfoPopupMessage } from '../../../store/appSlice';
 
 export default (props) => {
+  const deleteToken = useGetToken(); // fix
   const history = useHistory();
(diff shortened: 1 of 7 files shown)


simplq/src/components/pages/Status/LeaveQueue.jsx
```diff
@@ -2,22 +2,23 @@ import React from 'react';
...
   const dispatch = useDispatch();
-  const handleClick = () => {
-    props.leaveQueueHandler().then(() => {
-      dispatch(setInfoPopupMessage('Successfully left queue'));
-      history.push(`/`);
-    });
+  const onDeleteClick = () => {
+    dispatch(deleteToken(props.tokenId));
+    dispatch(setInfoPopupMessage('Successfully left queue'));
+    history.push(`/`);
```
daltonfury42, line 16: This is not fully correct. We should wait till the token is deleted and go back. I could add a comment and handle it later, but can't thing of a very elegant solution.

mradenovic: You can dispatch any action from async action. Just watch for circular dependencies.

```js
//  deleteToken.js
      try {
        dispatch(setInfoPopupMessage('Successfully left queue'));
        history.goBack();
        return response;
      } catch (error) {
        return rejectWithValue({ message: `Some message...` });
      }
```

PR #529, opened 2021-02-07
Load base url from config into api/

mradenovic:


daltonfury42:
This is duplicate of #525 

I did the same, plus did some eslint warning suppression and commenting

PR #531, opened 2021-02-08
Add LoadingStatus Component

mradenovic:
Upgrade storybook to 6.1.17

Fix LoadingIndicator stories

Add new Loading component

Use Loading wrapper in Join page

Separate Loading container from presentation

Add Loading stories

### DIFF
--- a/simplq/src/components/common/Loading/Loading.jsx
+++ b/simplq/src/components/common/Loading/Loading.jsx
@@ -1,0 +1,23 @@
+import React from 'react';
+import PropagateLoader from 'react-spinners/PropagateLoader';
+import styles from './Loading.module.scss';
+
+export default ({ children, actionStatus }) => {
+  if (!actionStatus || actionStatus === 'pending') {
+    return (
+      <div className={styles.main}>
+        <PropagateLoader color="#3a3768" />
+      </div>
+    );
+  }
+
+  if (!actionStatus || actionStatus === 'rejected') {
+    return (
+      <div className={styles.main}>
+        An unknown error occured. Please look at the console log for more nfo.
+      </div>
+    );
+  }
+
+  return <>{children}</>;
+};
--- a/simplq/src/components/pages/Join/index.jsx
+++ b/simplq/src/components/pages/Join/index.jsx
@@ -31,7 +31,7 @@ export default ({ history, match }) => {
     <div>
       <HeaderSection queueName={queueStatus.queueName} history={history} />
       <div className={styles['main-content']}>
-        <LoadingStatus dependsOn="getQueueStatusByName">
+        <LoadingStatus dependsOn={{ getQueueStatusByName }}>
           <div className={styles['queue-stats']}>
             <QueueStats queueStatus={queueStatus} />
           </div>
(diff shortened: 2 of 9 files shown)


simplq/src/components/pages/Join/index.jsx
```diff
@@ -35,15 +31,17 @@ export default ({ history, match }) => {
...
-        <div className={styles['queue-stats']}>
-          <QueueStats queueStatus={queueStatus} />
-        </div>
-        <p className={styles['message']}>Please enter your contact details to join this queue</p>
-        <JoinQueueForm
-          queueId={queueId}
-          joinQueueHandler={joinQueueHandler}
-          buttonText="Join Queue"
-        />
+        <LoadingStatus dependsOn="getQueueStatusByName">
```
daltonfury42, line 34: Should we also define `getQueueStatusByName` etc as constants somewhere and use it?

simplq/src/components/common/Loading/Loading.jsx
```diff
@@ -0,0 +1,23 @@
...
+      <div className={styles.main}>
+        <PropagateLoader color="#3a3768" />
+      </div>
+    );
+  }
+
+  if (!actionStatus || actionStatus === 'rejected') {
+    return (
+      <div className={styles.main}>
+        An unknown error occured. Please look at the console log for more nfo.
```
daltonfury42, line 17: Lets replace with this:

An unknown error occurred. Our team is looking into this. You can also reach out to us [here](https://github.com/SimplQ/simplQ-frontend/issues/new/choose) if you want to follow up on this.

daltonfury42: Also lets put this behaviour behind a flag `displayErrors` which would defaults to true (🤔 ) as it will otherwise upset tiny spinners.

daltonfury42 (commented):
This is a pretty neat idea.

daltonfury42:
We should also globally capture all API failures if possible and [send it](https://docs.sentry.io/product/sentry-basics/guides/integrate-backend/capturing-errors/) to sentry.

simplq/src/components/common/Loading/Loading.jsx
mradenovic: image

A bit too long 😄

> Also lets put this behaviour behind a flag displayErrors which would defaults to true (🤔 ) as it will otherwise upset tiny spinners.

I don't think anything is necessary here. It will just help to catch unhandled errors.

daltonfury42: Right. I think unhandled errors can be silently passed to sentry without polluting the spinner component.

simplq/src/components/pages/Join/index.jsx
mradenovic: I didn't like declaring, exporting, importing another variable. Just pass the whole action in an object.

```js
<LoadingStatus dependsOn={{ getQueueStatusByName }}>
```

Reads nicely: **Loading status depends on qetQueueStatusByName** 😄

navaneeth-spotnana (approved):
PropTypes is really cool!

daltonfury42 (approved):

PR #535, opened 2021-02-12
Move auth0 out of async actions.

mradenovic:
Since an async action is called whithin a hook, it is guaranteed
that it will be called within a component. It doesn't need to know
anything about auth0.


daltonfury42 (approved):
👍

PR #536, opened 2021-02-12
Add PropTypes to Loading component

mradenovic:


simplq/src/components/common/Loading/Loading.jsx
```diff
@@ -21,3 +22,15 @@ export default ({ children, actionStatus, isLoading }) => {
...
 };
+
+Loading.defaultProps = {
+  isLoading: undefined,
+  actionStatus: undefined,
+};
+
+Loading.propTypes = {
+  isLoading: PropTypes.bool,
+  actionStatus: PropTypes.oneOf(['pending', 'fulfilled', 'rejected']),
```
daltonfury42, line 33: Let us also add `children: PropTypes.element.isRequired`?

daltonfury42 (approved):

simplq/src/components/common/Loading/Loading.jsx
mradenovic: If we do that, the component can't be used on its own. E.g., dispatch joinQueue and bring modal spiner to the front until action is fulfilled.  Do we want to that?

daltonfury42: Ok, let's keep it as is. 👍

PR #539, opened 2021-02-13
Tour guide implementation: Implemented the suggested changes.

promarcussmith:
1) moved the stepChange function  into common/Tour/Tour.jsx and used it directly.

2) defined a index.jsx file in common/Tour.

3) Included the drop down element in the Copy Queue button during second step of tour.


promarcussmith:
Was caught up with other activities, so I was not able to look into the suggested changes. Today, I was able to work on this suggested changes by @daltonfury42 .

daltonfury42:
Can you run `npm run lint` and fix any issues that come up? Otherwise looks good to me.

@mradenovic should we have storybook stories for this as well?

promarcussmith:
> Can you run `npm run lint` and fix any issues that come up? Otherwise looks good to me.
> 
> @mradenovic should we have storybook stories for this as well?

I have already run npm run lint.

daltonfury42 (approved):

Issue #492, opened 2021-01-26
Streamline naming convention


mradenovic:
- All filenames except index.jsx should be in PascalCase (i.e. Join.jsx, Join.module.scss, Form.jsx)
- Use index.jsx to re-export classes from the folder
- Match route path with component name
```html
<Switch>
    <Route path="/queue/:queueId" exact component={AdminQueue} />
    <Route path="/queue/:queueId" exact component={QueueById} />

    <Route path="/j/:queueName" exact component={JoinQueueWithDetails} />
    <Route path="/j/:queueName" exact component={JoinQueueByName} />

    <Route path="/token/:tokenId" exact component={QueueStatus} />
    <Route path="/token/:tokenId" exact component={TokenById} />

    <Route path="/privacy" exact component={TermsOfService} />
    <Route path="/legal" exact component={Legal} /> <!-- Split PrivacyPolicy and TermsOfUse -->
</Switch>
```

Make small changes in the following order to avoid too much turbulence (pun intended 😄 ):
- [x] Components
- [x] Pages
- [x] Layout (matching should be done last)
- [x] Actions and request Factory

maaverik:
The only thing about this issue I have doubts about is 'Match route path with component name'. e.g., `Route path="/queue/:queueId"` points to the queue admin page (hence the component is named AdminQueue, we could shorten to Admin too), which requires a queueId to fetch info. Changing meaningful component names just to make it match the route name seems like overkill to me.

PR #544, opened 2021-02-18
Refactor home page

mradenovic:
* Rename files according to #492.
* Prevent a round trip to the server on every render of `MyQueues`


simplq/src/components/pages/Home/LandingPage.jsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 import { useAuth0 } from '@auth0/auth0-react';
 import Header from 'components/common/Header';
-import CreateJoinForm from 'components/common/CreateJoinForm';
-import styles from './home.module.scss';
+import QueueForm from 'components/common/CreateJoinForm';
+import styles from './Home.module.scss';
```
daltonfury42, line 5: should we have `Page` in the css file also? `HomePage.module.scss`

daltonfury42 (approved):

simplq/src/components/pages/Home/LandingPage.jsx
mradenovic: Every component should own its css.  That way it can be reused elsewhere. Right now `Home.module.scss` contains css for more than just `HomePage.jsx`. Once all separated, we might get rid of it completely.

PR #549, opened 2021-02-20
Refactoring and admin page redux refactor

maaverik:
Going slow and steady this time :)

PR #550, opened 2021-02-20
Refactor admin page

mradenovic:
- [x] `Token` manage token with dispatch. - @mradenovic 
- [x] `TokenList` read state from redux (selectTokens selector in selectedQueue). - @mradenovic 
- [x] `AdminPage` update redux store on load. - @mradenovic 
- [x] `AdminSidePanel` destructure props. - @maaverik
- [x] `DeleteQueue` delete queue with dispatch. - @maaverik 
- [x] `AddMember` add member with dispatch - @maaverik


maaverik:
I'll start with merging any missed changes from #549

mradenovic:
I made a small list of what needs to be done. Pick one component and work just on it.

maaverik:
I'll take DeleteQueue. It's late here so will work on it tomorrow.

simplq/src/components/pages/Admin/AdminPage.jsx
```diff
@@ -0,0 +1,66 @@
...
+  const [toursteps, setToursteps] = useState(getToursteps(window.innerHeight));
+  const { isAuthenticated } = useAuth0();
+
+  const update = useCallback(() => {
+    clearTimeout(timeoutId);
+    dispatch(getSelectedQueue({ queueId }));
+    setTimeout(update, TIMEOUT);
+    // TODO: Check if this is good solution.
+    /* eslint-disable-next-line */
+  }, [queueId]);
```
mradenovic, line 36: I'm not sure that this is the best solution, just experimenting.

mradenovic, line 36: I'm not sure that this is the best solution, just experimenting.

mradenovic:
Kinda ready. Most of the work is done with redux refactoring thanks to @maaverik and @daltonfury42 🍾 . There will be small things to iron out and a lot of `You Can Do This` issues 🤣

simplq/src/components/pages/Admin/AdminPage.jsx
```diff
@@ -0,0 +1,66 @@
...
+  const dispatch = useDispatch();
+  const getSelectedQueue = useGetSelectedQueue();
+
+  const [toursteps, setToursteps] = useState(getToursteps(window.innerHeight));
+  const { isAuthenticated } = useAuth0();
+
+  const update = useCallback(() => {
+    clearTimeout(timeoutId);
+    dispatch(getSelectedQueue({ queueId }));
+    setTimeout(update, TIMEOUT);
```
chalx, line 33: you need to assign the return to `timeoutId` but I don't really understand what are you trying to do here.

simplq/src/components/pages/Admin/AdminPage.jsx
```diff
@@ -0,0 +1,66 @@
...
+  const resize = () => setToursteps(getToursteps(window.innerWidth));
+
+  useEffect(() => {
+    window.addEventListener('resize', resize);
+    return () => window.removeEventListener('resize', resize);
+  });
+
+  useEffect(() => {
+    update();
+    return () => clearTimeout(timeoutId);
```
chalx, line 47: @mradenovic here `timeoutId` is not set

simplq/src/components/pages/Admin/TokenList.jsx
```diff
@@ -1,40 +1,23 @@
...
       <div className={styles['token-list']}>
-        {props.tokens.map((token) => (
-          <Token
-            removeTokenHandler={props.removeTokenHandler}
-            token={token}
-            queueId={queueId}
-            key={token.tokenId}
-          />
-        ))}
+        <Dispaly />
```
maaverik, line 17: Typo. Maybe call it Cards or something else?

simplq/src/store/asyncActions/deleteToken.js
```diff
@@ -21,8 +21,9 @@ const useDeleteToken = () => {
     const { tokenId } = arg;
     const authedRequest = makeAuthedRequest(RequestFactory.deleteToken(tokenId));
     const response = await authedRequest;
-    dispatch(setInfoPopupMessage('Successfully left queue'));
+    // TODO: remove token from redux store on success
     if (arg.goHome) {
```
maaverik, line 25: After adding logic for this and similar logic for adding members to modify the token slice, the page should be much more responsive. Let's do that in a subsequent PR.

maaverik (commented):
This almost brings our redux migration to a close. I was able to learn a lot about redux in the process :smile:. Kudos @mradenovic.

simplq/src/components/pages/Admin/AdminPage.jsx
maaverik: Best to fix this. Although it doesn't fail, it could cause a memory leak I believe. Later on, we could make something like [this](https://overreacted.io/making-setinterval-declarative-with-react-hooks/#just-show-me-the-code). But for now, let's just fix and use this @mradenovic.

simplq/src/components/pages/Admin/TokenList.jsx
mradenovic: I'm open to suggestions. `ToListOrNotToList` 🤣

simplq/src/components/pages/Admin/AdminPage.jsx
mradenovic: Good catch @chalx. Thanks!

> Later on, we could make something like this. But for now, let's just fix and use this @mradenovic.

I like it! Any timer functionality should be moved out of async action (e.g., `../asyncActions/getTotken.jsx`). Then async action can be reused. There is probably a better way to do it than what I did her.

simplq/src/components/pages/Admin/TokenList.jsx
mradenovic: `ListContent`?

maaverik: Yeah, that works. I'm thinking if we should move all the token logic into this file. That way, if a token is added or removed, only the token list rerenders instead of the whole admin page. Do you see any problems with that?

mradenovic: Both header and token list depend on `getSelectedQueue`. Since only the redux store is changing, and only the tokens field is changing, I believe only the token list will rerender.

maaverik: Okay, got it

maaverik (approved):

simplq/src/components/pages/Admin/TokenList.jsx
mradenovic: I'm wrong, Admin page renders every time.

maaverik: Yes, I just checked with some logs, seems like the admin component does rerender everytime. I'll make this change soon.

PR #553, opened 2021-02-21
Changed createJoinForm to use redux actions

maaverik:


simplq/src/store/asyncActions/joinQueue.js
```diff
@@ -21,7 +21,9 @@ const useJoinQueue = () => {
         RequestFactory.createToken(name, contactNumber, notifiable, queueId)
       );
       const response = await authedRequest;
-      history.push(`/token/${response.tokenId}`);
+      if (response) {
+        history.push(`/token/${response.tokenId}`);
+      }
       return response;
```
maaverik, line 27: I'm not sure if this condition is required, redux seems to implicitly handle cases where this request fails.

simplq/src/components/common/CreateJoinForm/CreateJoinForm.jsx
```diff
@@ -61,18 +56,14 @@ const CreateJoinForm = (props) => {
...
-          <>
-            <div>
-              <StandardButton onClick={handleCreateClick}>Create Queue</StandardButton>
-            </div>
-            <div>
-              <StandardButton onClick={handleJoinClick}>Join Queue</StandardButton>
-            </div>
-          </>
-        )}
+        <Loading isLoading={actionStatus === 'pending'}>
```
mradenovic, line 59: `LoadingStatus` component solves this common problem. One less import,  one less selector.

```jsx
<LoadingStatus dependsOn="createQueue">Content</LoadingStatus>
```

daltonfury42, line 59: You can use `<LoadingStatus />` here.  Check [this](https://github.com/SimplQ/simplQ-frontend/blob/master/simplq/src/components/pages/Join/JoinPage.jsx#L35) out.

simplq/src/store/asyncActions/joinQueue.js
daltonfury42: I am also not sure, I'd vote to remove the condition and fix later if there is an issue.

simplq/src/components/common/CreateJoinForm/CreateJoinForm.jsx
```diff
@@ -1,33 +1,28 @@
 import React, { useState } from 'react';
```
daltonfury42, line 1: Can we also add some stories while we are at it?

simplq/src/store/asyncActions/joinQueue.js
mradenovic: [createAsyncThunk](https://redux-toolkit.js.org/api/createAsyncThunk#return-value) always resolves, and dispatches `pending` action first, and then `fulfilled` or `rejected` action. `try catch` is needed only if you want to handle rejected promise here.

maaverik: We'll need to add try-catches at a few places including this one. We used to display API failures using [handleApiErrors](https://github.com/SimplQ/simplQ-frontend/blob/7e8c77088cf18ea66a0de918c127edbc49281ace/simplq/src/api/axios.js#L8), but this doesn't work with our redux setup. So we'll either have to use this function in [makeAuthedRequest](https://github.com/SimplQ/simplQ-frontend/blob/7e8c77088cf18ea66a0de918c127edbc49281ace/simplq/src/api/auth.js#L23) or add try-catches and extraReducers to AppSlice wherever required.

simplq/src/components/common/CreateJoinForm/CreateJoinForm.jsx
maaverik: LoadingStatus doesn't work here in its current form. We need the buttons to be displayed at the start and only show the Loading component after click, but since the CreateQueue action isn't triggered on page load, the corresponding actionStatus resolves to undefined, so the Loading component is always displayed instead of the children. We'll need to handle this case in LoadingStatus to be able to use it better.

maaverik: https://github.com/SimplQ/simplQ-frontend/pull/553/files#r579805848

daltonfury42: Got it. It makes sense to spin only on 'pending' and `isLoading=true` [here](https://github.com/SimplQ/simplQ-frontend/blob/master/simplq/src/components/common/Loading/Loading.jsx). 
```
if (actionStatus === 'rejected') { do current behavious, show error }
if (actionStatus === 'pending' || isLoading === true) { spin }
else render children
```

maaverik: Yes, I think this will not cause any problems. I'll make the changes

mradenovic (approved):
I'm gonna merge this as-is, and rename and make stories later.

simplq/src/store/asyncActions/joinQueue.js
daltonfury42: The actionStatusSlice way is a better approach.

PR #557, opened 2021-02-26
Fix security vulnerabilities

hallelshohat:


simplq/public/index.html
```diff
@@ -57,7 +57,7 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>SimplQ</title>
-    <script async defer src="https://buttons.github.io/buttons.js"></script>
+    <script async defer src="https://buttons.github.io/buttons.js" integrity="sha384-wwvJUp8hgwWMT+bfKNmxsQuctQq/ACKlkQYTpgRaQQdVXsBy9jfUA8a1h6GFWcl8" crossorigin="anonymous"></script>
```
daltonfury42, line 60: How did you get the sha?

daltonfury42 (commented):
Thanks for the PR

simplq/public/index.html
hallelshohat: the hash ensures that the resource itself will not change, and if the resource will change it will break the site.
I will remove it here

daltonfury42 (approved):

PR #559, opened 2021-03-01
Minor Changes

JJ-Hall:
changed some file names and removed commits based on SonarCloud


simplq/src/components/pages/Admin/AdminPage.jsx
```diff
@@ -7,7 +7,7 @@ import { selectQueueName } from 'store/selectedQueue';
 import Ribbon from 'components/common/Ribbon';
 import Tour from 'components/common/Tour';
 import { useGetSelectedQueue } from 'store/asyncActions';
-import HeaderSection from './AdminHeaderSection';
+import HeaderSection from './HeaderSection';
```
mradenovic, line 10: > By convention, a file that exports only one class, function, or constant should be named for that class, function or constant. Anything else may confuse maintainers.

In this case, the class name should be changed, not the file name.

simplq/src/components/Layout/Routes.jsx
```diff
@@ -5,7 +5,7 @@ import QueueStatus from 'components/pages/Status';
 import AdminPage from 'components/pages/Admin/AdminPage';
 import PageNotFound from 'components/pages/PageNotFound';
 import PopupNotifications from 'components/common/Popup';
-import HomePage from 'components/pages/Home/HomePage';
+import HomePage from 'components/pages/Home/Home';
```
mradenovic, line 8: > By convention, a file that exports only one class, function, or constant should be named for that class, function or constant. Anything else may confuse maintainers.

In this case, the class name should be changed, not the file name.

mradenovic (commented):
@JJ-Hall, thanks for trying! Wherever the reason for the code smell rename suggestion is the single class export, you should opt to change the name of the class, rather than the filename. The most common reason for this case is that developer (including myself 😊 ) decided to change the file name for some good reason but forgot to change the class name. The second most common is creating files by copy-paste.

simplq/src/components/pages/Admin/AdminPage.jsx
JJ-Hall: @mradenovic I thought it was funky that it was having me change the file name. I appreciate the explanation.

daltonfury42:
@JJ-Hall can you merge latest master into your branch, and let's get this merged.

Issue #548, opened 2021-02-20
Fix Sonarqube reported Code Smells


daltonfury42:
We have accumulated some easy to fix security vulnerabilities over the last few months over [here](https://sonarcloud.io/project/security_hotspots?id=SimplQ_simplQ-frontend&branch=master&resolved=false&types=SECURITY_HOTSPOT#). 

These are straightforward easy fixes, good for a new comer to take up as a good first issue. 

Once they are fixed, we should see if a branch protection rule can be set to catch them pre-merge

hallelshohat:
I've made a pull request solving this issue

daltonfury42:
Thanks @hallelshohat for fixing the security hotspots. Now what is left is some code smells that if anyone is interested, can go through and fix.

https://github.com/SimplQ/simplQ-frontend/pull/557

daltonfury42:
I have gone through the list of issues and resolved a few as false positive after our discussions on https://github.com/SimplQ/simplQ-frontend/pull/564

@maaverik @mradenovic If you get time, go through the codesmells and let me know if we need there are any more such false positives

skShekhar:
@daltonfury42 Are there anymore code smells or can this issue be closed?

maaverik:
Hey @skShekhar. Sorry, missed your comment here. There are still a few small issues left, like the many TODOs we have. Let's leave this open so that we at least have a tracker that we can look at.

Mogakamo:
Hey @daltonfury42 I'm new to open source and I wanted to tackle this issue.
I was hoping you could give me a heads up on how I could tackle this one.

daltonfury42:
Hi @Mogakamo welcome. Here is a list of tiny tech debts that you can solve:

https://sonarcloud.io/project/issues?id=SimplQ_simplQ-frontend&open=AXfKRRGH1JV0IME_FJy0&resolved=false
https://sonarcloud.io/project/issues?id=SimplQ_simplQ-frontend&open=AXhz9D02G5SMudjmecKx&resolved=false
https://sonarcloud.io/project/issues?id=SimplQ_simplQ-frontend&open=AXe6u56LmUtjMyP-a6uc&resolved=false

After that, you can take up a bigger task like #601 

#643 is an even bigger project would require you to first do the backend APIs also.

Mogakamo:
ok thanks for the heads up @daltonfury42. I'll get back when i'm done

divyansh21jain:
[...]

PR #563, opened 2021-03-03
Fix return statement to immediately return data

kabaliisa:


daltonfury42 (approved):
Good one! Please send more PRs as you find things like these.

simplq/src/api/auth.js
```diff
@@ -52,8 +52,8 @@ const useMakeAuthedRequest = () => {
         Authorization: await getAuthHeaderValue(auth),
       },
     });
-    const response = await authedRequest;
-    return response.data;
+
+    return data;
   };
```
maaverik, line 57: I was under the impression that these two versions are functionally equivalent. Could someone help me understand what I'm missing here 😕 ?

kabaliisa: > I was under the impression that these two versions are functionally equivalent. Could someone help me understand what I'm missing here  ?

@maaverik using less variables reduces complexity of the algorithm. so instead of having 2 variables authedRequest and response we an have one.

kabaliisa: Btw I'm getting this list of issues from [https://sonarcloud.io/project/issues?id=SimplQ_simplQ-frontend&open=AXe6u6A1mUtjMyP-a6uv&resolved=false&severities=MINOR&sinceLeakPeriod=true&types=CODE_SMELL](url) I hope it's Ok

maaverik: > @maaverik using less variables reduces complexity of the algorithm. so instead of having 2 variables authedRequest and response we an have one.

Okay, got it. I thought there was something with promise returns that I was missing. Using less variables is a good point.

maaverik: > 
> 
> Btw I'm getting this list of issues from [here](https://github.com/SimplQ/simplQ-frontend/issues/548#issuecomment-787064042) .I hope it's Ok

Please go ahead, these are all things we should have fixed before.

daltonfury42: @maaverik These are all code smells returned by Sonarqube.

@kabaliisa While you fix the issues, can you add a link to the corresponding sonarqube page?

daltonfury42: These are great change and we are also learning a lot from your PRs. :)

kabaliisa: > @maaverik These are all code smells returned by Sonarqube.
> 
> @kabaliisa While you fix the issues, can you add a link to the corresponding sonarqube page in the PR description?

Cool

mradenovic: This code smell was about `return makeAuthedRequest` and I am pretty sure it will still `smell` after this PR.

daltonfury42: This change looks good to me. @mradenovic 

```
const { data } = await ...
return data;
```

is better than 
```
authedRequest = ...
response = await ...
return response.data
```

The smell about `return makeAuthedRequest`, let me check and suppress it if it's not making sense.

PR #564, opened 2021-03-04
Fix return statement to immediately  return expression

kabaliisa:


maaverik:
This is a purely personal preference, but I like the old version 😅. It just seems clearer that directly returning a function. But @daltonfury42 @mradenovic if you think this version is better, we can go with it.

daltonfury42:
These are all subjective, but generally I avoid creating an unwanted reference unless it's really required. I've seen it followed in  my team at work as well.

It takes a while to get used to, but once you adapt this style, you will see that it's nice. Code will be more lean and looking functional etc. And less buggy I've observed.

If you look at our backend code as well, you will references used only when necessary.

daltonfury42:
There are more tiny things like these that we have not strictly followed, for example we could have used object [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) more.

daltonfury42 (approved):

mradenovic:
Late to the party, but I preferred the old version too. It is more obvious what is returned. And readability should always have priority over style.

If we just want to fix `code smell`, we could do ` return { createQueue }`.

daltonfury42:
> Late to the party, but I preferred the old version too. It is more obvious what is returned. And readability should always have priority over style.

Alright, two of you have voted, so let's revert. Readability wins. I will silence this codesmell on Sonarqube, and revert this and any other same change tomorrow.

> If we just want to fix `code smell`, we could do ` return { createQueue }`.

Nah, let's just revert and suppress the code smell.

cc @kabaliisa Let's not do this for anymore return statements from actions.

kabaliisa:
> > Late to the party, but I preferred the old version too. It is more obvious what is returned. And readability should always have priority over style.
> 
> Alright, two of you have voted, so let's revert. Readability wins. I will silence this codesmell on Sonarqube, and revert this and any other same change tomorrow.
> 
> > If we just want to fix `code smell`, we could do ` return { createQueue }`.
> 
> Nah, let's just revert and suppress the code smell.
> 
> cc @kabaliisa Let's not do this for anymore return statements from actions.

Alright!

PR #565, opened 2021-03-04
change: rename files

kabaliisa:


maaverik:
@mradenovic has already addressed changing filenames for HomePage and HeaderSection in #559.  

As for TourSteps, I feel that it's more consistent to keep the old filename, but change the component name in the file to TourSteps for consistency. Even though it is returning an array and not a component, it has JSX in it, so giving it a functional name with 'get' in it doesn't seem right.

kabaliisa:
> @mradenovic has already addressed changing filenames for HomePage and HeaderSection in #559.
> 
> As for TourSteps, I feel that it's more consistent to keep the old filename, but change the component name in the file to TourSteps. Even though it is returning an array and not a component, it has JSX in it, so giving it a functional name with 'get' in it doesn't seem right. But I'm no expert on naming conventions, so please correct me if I'm going wrong somewhere  .

Oh! looks like this was handled. You are right about  the 

> @mradenovic has already addressed changing filenames for HomePage and HeaderSection in #559.
> 
> As for TourSteps, I feel that it's more consistent to keep the old filename, but change the component name in the file to TourSteps. Even though it is returning an array and not a component, it has JSX in it, so giving it a functional name with 'get' in it doesn't seem right. But I'm no expert on naming conventions, so please correct me if I'm going wrong somewhere  .

Oh! Looks most these were handled. You may be right about the getTourSteps component. All components are Capitalized getTourSteps breaks the convention @daltonfury42

daltonfury42:
`getTourSteps` is not a component, so let us leave it as `getTourSteps` itself. https://github.com/SimplQ/simplQ-frontend/pull/559's author has not responded yet, so lets change this PR as per @mradenovic's suggestions and merge this.

@maaverik The file extension `.jsx` is to do with [Babel](https://babeljs.io/docs/en/) than React/Components. A simple rule that can be followed is, whenever the a JS file has HTML like stuff inside it, use `.jsx`

simplq/src/components/Layout/Routes.jsx
```diff
@@ -5,7 +5,7 @@ import TokenStatusPage from 'components/pages/TokenStatus/TokenStatusPage';
 import AdminPage from 'components/pages/Admin/AdminPage';
 import PageNotFound from 'components/pages/PageNotFound';
 import PopupNotifications from 'components/common/Popup';
-import HomePage from 'components/pages/Home/HomePage';
+import Home from 'components/pages/Home/HomePage';
```
mradenovic, line 8: I have addressed filename changes in #559. As explained [here](https://github.com/SimplQ/simplQ-frontend/pull/559#pullrequestreview-600687938):

> Wherever the reason for the code smell rename suggestion is the single class export, you should opt to change the name of the class, rather than the filename. The most common reason for this case is that developer (including myself 😊 ) decided to change the file name for some good reason but forgot to change the class name. The second most common is creating files by copy-paste.

mradenovic (changes requested):

simplq/src/components/Layout/Routes.jsx
kabaliisa: > I have addressed filename changes in #559. As explained [here](https://github.com/SimplQ/simplQ-frontend/pull/559#pullrequestreview-600687938):
> 
> > Wherever the reason for the code smell rename suggestion is the single class export, you should opt to change the name of the class, rather than the filename. The most common reason for this case is that developer (including myself blush ) decided to change the file name for some good reason but forgot to change the class name. The second most common is creating files by copy-paste.

@mradenovic  should I revert this ?

mradenovic: Yes please! You can change the class name in `HomePage.tsx` to match the filename.

daltonfury42 (approved):
LGTM

daltonfury42:
@kabaliisa Thanks for your PR.

kabaliisa:
> @kabaliisa Thanks for your PR.
> 
> Edit: I think @mradenovic needs to approve...

You're welcome @daltonfury42

mradenovic (approved):

mradenovic:
> @kabaliisa Thanks for your PR.
> 
> Edit: I think @mradenovic needs to approve...

Sorry for the delay!

Issue #21, opened 2020-04-22
Get a good domain name


daltonfury42:

daltonfury42:
https://simplq.me it is

PR #574, opened 2021-03-08
smell issue#21- return authedRequest

ME-ON1:
As suggested by the smell issue [#21](https://sonarcloud.io/project/issues?id=SimplQ_simplQ-frontend&open=AXe6u6BMmUtjMyP-a6uy&resolved=false&sinceLeakPeriod=true&types=CODE_SMELL)

### DIFF
--- a/simplq/src/store/asyncActions/getUserQueues.js
+++ b/simplq/src/store/asyncActions/getUserQueues.js
@@ -19,9 +19,7 @@ const useGetUserQueues = () => {
   const makeAuthedRequest = useMakeAuthedRequest();
 
   const getUserQueues = createAsyncThunk(typePrefix, async () => {
-    const authedRequest = makeAuthedRequest(RequestFactory.getUserQueues());
-    const response = await authedRequest;
-    return response;
+    return makeAuthedRequest(RequestFactory.getUserQueues());
   });
 
   return getUserQueues;
--- a/simplq/src/store/asyncActions/notifyToken.js
+++ b/simplq/src/store/asyncActions/notifyToken.js
@@ -13,9 +13,7 @@ const useNotifyToken = () => {
   const makeAuthedRequest = useMakeAuthedRequest();
 
   const notifyToken = createAsyncThunk(typePrefix, async ({ tokenId }) => {
-    const authedRequest = makeAuthedRequest(RequestFactory.notifyToken(tokenId));
-    const response = await authedRequest;
-    return response;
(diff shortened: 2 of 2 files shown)


simplq/src/store/asyncActions/getUserQueues.js
```diff
@@ -19,9 +19,8 @@ const useGetUserQueues = () => {
   const makeAuthedRequest = useMakeAuthedRequest();
 
   const getUserQueues = createAsyncThunk(typePrefix, async () => {
-    const authedRequest = makeAuthedRequest(RequestFactory.getUserQueues());
-    const response = await authedRequest;
-    return response;
+    const authedRequest = await makeAuthedRequest(RequestFactory.getUserQueues());
+    return authedRequest;
```
daltonfury42, line 23: Let's do this in one line:
```
return await makeAuthedRequest(RequestFactory.getUserQueues());
```

daltonfury42 (changes requested):

simplq/src/store/asyncActions/getUserQueues.js
ME-ON1: Ohh right,  On it !!

ME-ON1: Done please check that again !

daltonfury42:
Looks like you missed an await

ME-ON1:
NO, i removed that purposefully. 
`return await makeAuthedRequest(RequestFactory.getUserQueues());` 
wasn't possible as it was causing error by eLint  
`error  Redundant use of "await" on a return value  no-return-await` 
and after some research i found that 
`return await fun()` is same as `return  fun()` 
inside a async function.

daltonfury42 (approved):
Nice. Didn't know this.

maaverik:
> and after some research i found that
> `return await fun()` is same as `return fun()`
> inside a async function.
Hi @ME-ON1, thanks for the PR. Just wanted to double-check if you sure that they are equivalent. This [MDN doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) at the end says that they are not equivalent and if so, I think we should retain await in the function because we are not awaiting anywhere outside the thunk. Could you please check again if this change would cause any problems?

daltonfury42:
They are not equivalent in the sense that the function doesn't block. It immediately returns as a promise, and hence is little more performant.

The difference is very subtle, we would need to await only if say we want to block and catch errors inside the function etc. Since we don't want to do that, this should work for us.

PR #582, opened 2021-03-20
Handle token add/remove in selectedQueueSlice

daltonfury42:
Resolves https://github.com/SimplQ/simplQ-frontend/issues/577

PR #583, opened 2021-03-20
Centralised request updates and failures

daltonfury42:
1. Centralised request updates and failures
2. Remove unused old file makeRequest.js


simplq/src/api/auth.js
```diff
@@ -51,6 +52,18 @@ const useMakeAuthedRequest = () => {
...
       },
+    }).catch((error) => {
+      // log error to sentry for alerting
+      Sentry.captureException(error);
+      // In case of request failure, extract error from response body
+      if (error.response) {
+        // Response has been received from the server
+        const message = error.response.data.message;
+        throw new Error(message || 'Unknow error occured');
+      } else {
```
maaverik, line 63: Should we add this?
`else if (err.response.status === 422) {
    store.dispatch(
      setErrorPopupMessage(
        There's a problem with the data you've entered ${err.response.data.message}
      )
    );
  }`

I remember we had added this in handleApiErrors because the default error message in this case was kind of awkward for this particular kind of error.

maaverik, line 63: Typo for unknown

simplq/src/store/appSlice.js
```diff
@@ -20,28 +24,30 @@ const appSlice = createSlice({
...
-    [joinQueue.rejected]: (state, action) => {
-      state.errorText = action.error.message;
-    },
-    [joinQueue.fulfilled]: (state, action) => {
-      state.infoText = `Added to ${action.payload.queueName}`;
-    },
+  extraReducers: (builder) => {
+    builder
+      .addCase(joinQueue.pending, (state) => {
+        state.infoText = `Adding to queue...`;
```
maaverik, line 30: What's the reason for using builder.addCase?? Is it somehow better than the old version?

daltonfury42: Yes, it's the recommended way, plus you can do advanced matching, see the last case in the chain.

https://redux-toolkit.js.org/api/createSlice#the-extrareducers-builder-callback-notation

maaverik: Okay, got it :+1:

simplq/src/api/auth.js
daltonfury42: What current logic does is,

```
A) If there is no response, it's a network issue, ask user to check internet.

B) If response body has a "message", just show it to the user. Backend sends user presentable messages like this.

C) Else, we don't really know, so "Unknown error occurred" pops up.
```

`handleApiErrors` is also printing the same `error.response.data.message` as in B, with a prefix. But now with redux, we can decouple the reaction to the event and the originating of the event. (The beauty of actions)

daltonfury42: This would slightly change if we ever support multiple languages, but the overall architecture LGTM.

I'll take @mradenovic 's views on this and https://github.com/SimplQ/simplQ-frontend/pull/582 before we merge.

maaverik: Yeah, I understand. I think we added that prefix because the messages the server sent back in case of 422 errors used to not be very obvious to users, but it might have been fixed later on. If so, it's not required.

daltonfury42: Let's fix them at the backend side, when you notice such exceptions, flag them, and we will handle it at backend side.

https://github.com/SimplQ/simplQ-backend/blob/master/simplq/src/main/java/me/simplq/exceptions/SQInvalidRequestException.java

Also, we should hopefully start logging all failures to sentry with this change, so we can refine the error text over time as we come across them.

maaverik: Once this is merged, let's move all the changes so far to Prod. We could finally do a 1.0 release.

daltonfury42: yes, right. I'll merge on your approval. 

Then lets test dev to see if everything is working, and we can go 1.0 ✌️

maaverik (approved):

PR #591, opened 2021-03-23
Show spiner on first load in token status page

### LINKED ISSUE #585: Prevent page shake on background refresh
On the token status page, we periodically call the backend every 10 seconds to check for updates. While this happens, the page gets a loader, and it also shakes up a bit because the loader reduces the size of components.

https://user-images.githubusercontent.com/11256376/111901030-aaf87a00-8a5b-11eb-8cf6-8925fb906965.mov

Wondering what is the best way to handle this. I could think of two ways:

1. Use CSS to make sure the sizes don't change when loader is rendered vs the contect.
2. Show loader only on first load, don't render on subsequent refreshes. 

Thoughts?

mradenovic:
Resolves #585

### DIFF
--- a/simplq/src/components/pages/TokenStatus/TokenNumber.jsx
+++ b/simplq/src/components/pages/TokenStatus/TokenNumber.jsx
@@ -1,4 +1,4 @@
-import LoadingStatus from 'components/common/Loading';
+import Loading from 'components/common/Loading/Loading';
 import React from 'react';
 import { useSelector } from 'react-redux';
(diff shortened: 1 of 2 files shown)


mradenovic:
Since `token` is guaranteed to be at least `{}` by redux, if the object is empty, `token` is still loading. Action status can be used if some indicator of subsequent loads needs to be displayed.

daltonfury42 (approved):
This works!

maaverik (approved):

PR #592, opened 2021-03-23
Fix Shaky Sidepanel

daltonfury42:
When a request goes pending, we were setting state to `{}`. This is not required. It makes the data go empty while request is going on in the queue stats on admin page, token page etc.

While at it, I also moved the code to get the queue stats from store down into the stats component where it's required.

After this PR, the page doesn't shake anywhere, hopefully.

### DIFF
--- a/simplq/src/components/common/QueueStats/QueueStats.jsx
+++ b/simplq/src/components/common/QueueStats/QueueStats.jsx
@@ -1,23 +1,34 @@
-import React, { useMemo } from 'react';
+import React, { useEffect, useCallback, useMemo } from 'react';
 import moment from 'moment';
+import { useGetQueueStatus } from 'store/asyncActions';
+import { selectQueueStatus } from 'store/queueStatus';
+import { useDispatch, useSelector } from 'react-redux';
+import { selectTokens } from 'store/selectedQueue';
 import styles from './QueueStats.module.scss';
 
-const DetailRow = ({ title, value, large }) => (
-  <div className={styles['detail-row']}>
-    <span className={styles['detail-name']}>{title}</span>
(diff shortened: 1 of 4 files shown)


simplq/src/components/common/QueueStats/QueueStats.jsx
```diff
@@ -1,5 +1,9 @@
-import React, { useMemo } from 'react';
+import React, { useEffect, useCallback, useMemo } from 'react';
 import moment from 'moment';
+import { useGetQueueStatus } from 'store/asyncActions';
+import { selectQueueStatus } from 'store/queueStatus';
+import { useDispatch, useSelector } from 'react-redux';
```
maaverik, line 5: Maybe we should rename selectQueueStatus to something like selectQueueDetails. In the pauseQueue PR, we have a selector with the same name and it makes more sense there.

maaverik: Also useGetQueueStatus to useGetQueueDetails since we have useSetQueueStatus coming in that PR too.

daltonfury42: Yeah, right. What to do think about "QueueInfo" here and for corresponding API and slices and component for this?

This API gives "info" about the queue, the components shows queue "information" etc..

Stats -> Info

maaverik: Yeah, sounds good

daltonfury42: I've handled it as a separate PR here: https://github.com/SimplQ/simplQ-frontend/pull/593

Issue #425, opened 2020-12-12
Add Configurable Max Queue Capacity Limit


daltonfury42:
Admin can optionally configure a max capacity, after which new people will be put on hold.

This might add value in certain situations, but I am a bit reluctant, as it's not a core feature. For example, admin can also adjust without it, I feel.

daltonfury42:
We should have a new right side panel button to open a settings dialogue box, in which we can add this as a optionally configurable entry.

Right now the modal would look empty with just one configuration item, but soon we will add more.

![image](https://user-images.githubusercontent.com/11256376/111103509-a3753480-8574-11eb-820e-6926ad7211c7.png)


To set the max queue capacity, send a patch request to `/queue/{queueId}`, example payload:

```
{
  "maxQueueCapacity": <some-number>
}
```

Currently we have not handled the case at backend side where we need it to be unlimited, so we can set a very high value say 100000 for now when the user un sets it.

maaverik:
Closed by #603

PR #599, opened 2021-03-28
Fix Sonarqube code smells and lint errors

skShekhar:
This PR fixes the left over Sonar code [smells](https://sonarcloud.io/project/issues?id=SimplQ_simplQ-frontend&resolved=false&sinceLeakPeriod=true&types=CODE_SMELL) mentioned in #548.

- Changed `paths` to `moduleDirectory` in `import/resolver` to fix the `Unable to resolve path to module xyz` eslint warning

### DIFF
--- a/simplq/.eslintrc.json
+++ b/simplq/.eslintrc.json
@@ -9,7 +9,7 @@
   "settings": {
     "import/resolver": {
       "node": {
-        "paths": ["src"]
+        "moduleDirectory": ["src", "node_modules"]
       }
     }
   },
--- a/simplq/.eslintrc.json
+++ b/simplq/.eslintrc.json
@@ -9,7 +9,7 @@
   "settings": {
     "import/resolver": {
       "node": {
-        "moduleDirectory": ["src", "node_modules"]
+        "paths": ["src"]
       }
     }
   },
--- a/simplq/src/store/asyncActions/getQueueInfo.js
+++ b/simplq/src/store/asyncActions/getQueueInfo.js
@@ -12,9 +12,11 @@ const typePrefix = 'getQueueInfo/action';
 const useGetQueueInfo = () => {
   const makeAuthedRequest = useMakeAuthedRequest();
 
-  return createAsyncThunk(typePrefix, async ({ queueId }) =>
+  const getQueueInfo = createAsyncThunk(typePrefix, async ({ queueId }) =>
     makeAuthedRequest(RequestFactory.getQueueInfo(queueId))
   );
+
+  return getQueueInfo;
 };
 
 const getQueueInfo = createAsyncThunk(typePrefix);
--- a/simplq/src/store/asyncActions/getQueueInfo.js
+++ b/simplq/src/store/asyncActions/getQueueInfo.js
@@ -12,13 +12,9 @@ const typePrefix = 'getQueueInfo/action';
 const useGetQueueInfo = () => {
   const makeAuthedRequest = useMakeAuthedRequest();
 
-  const getQueueInfo = createAsyncThunk(typePrefix, async ({ queueId }) => {
-    const authedRequest = makeAuthedRequest(RequestFactory.getQueueInfo(queueId));
-    const response = await authedRequest;
-    return response;
-  });
-
-  return getQueueInfo;
+  return createAsyncThunk(typePrefix, async ({ queueId }) =>
+    makeAuthedRequest(RequestFactory.getQueueInfo(queueId))
+  );
 };
 
 const getQueueInfo = createAsyncThunk(typePrefix);
--- a/simplq/src/store/asyncActions/getQueueInfoByName.js
+++ b/simplq/src/store/asyncActions/getQueueInfoByName.js
@@ -14,17 +14,22 @@ const useGetQueueInfoByName = () => {
   const makeAuthedRequest = useMakeAuthedRequest();
   const history = useHistory();
 
-  return createAsyncThunk(typePrefix, async ({ queueName }, { rejectWithValue }) => {
-    const authedRequest = makeAuthedRequest(RequestFactory.getQueueInfoByName(queueName));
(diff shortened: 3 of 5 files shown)


simplq/src/store/asyncActions/getQueueInfo.js
```diff
@@ -12,13 +12,9 @@ const typePrefix = 'getQueueInfo/action';
...
-  const getQueueInfo = createAsyncThunk(typePrefix, async ({ queueId }) => {
-    const authedRequest = makeAuthedRequest(RequestFactory.getQueueInfo(queueId));
-    const response = await authedRequest;
-    return response;
-  });
-
-  return getQueueInfo;
+  return createAsyncThunk(typePrefix, async ({ queueId }) =>
+    makeAuthedRequest(RequestFactory.getQueueInfo(queueId))
+  );
```
maaverik, line 17: Removing the return response here is a problem I believe. If you look at the queueInfoSlice script in the store directory, you'll see that we have a reducer that stores the response from this thunk into the state which is displayed in the admin or join pages.
```
    [getQueueInfo.fulfilled]: (state, action) => {
      return action.payload;
    },
```
So this change would need to be reverted.

maaverik: Apart from returning the API response body, let's revert the changes in this file and the following asyncActions files. The reason is that even though directly returning the thunks without naming them is arguably a better practice according to sonarcloud, we had decided before to leave the asyncActions thunks the way they were because we felt that it was a little more readable. It's a subjective choice, but we're going to stick to it. For the remaining asyncActions files, we had removed the related warnings from sonarcloud, but I guess we missed these since they were recently modified. Sorry about that. @daltonfury42 could you please remove the unnecessary smells from sonarcloud?

maaverik (changes requested):

simplq/src/store/asyncActions/getQueueInfo.js
navaneeth-spotnana: There are two separate things here:
A:

> Removing the return response here is a problem I believe. 

@maaverik he is using lambda without braces, so the expression becomes the return value:
```
(a, b) => a + b
``` 
vs 
```
(a, b) => {
  return a + b;
}
```
where the former is preferable. 

B: In the actions, we have decided to prefer using this form for readability:
```
  const getUserQueues = createAsyncThunk(typePrefix, async () => {
    ...
  });

  return getUserQueues;
```

over returning directly like this:
```
  return createAsyncThunk(typePrefix, async () => {
    ...
  });
```

So combining these two things, what we should do here is:

```
const useGetQueueInfo = () => {
  const makeAuthedRequest = useMakeAuthedRequest();
  
  const getQueueInfo = createAsyncThunk(typePrefix, async ({ queueId }) =>
    makeAuthedRequest(RequestFactory.getQueueInfo(queueId))
  );

  return getQueueInfo; 
}
```

> could you please remove the unnecessary smell warnings from sonarcloud?

On it.

simplq/src/store/asyncActions/getQueueInfoByName.js
```diff
@@ -14,22 +14,17 @@ const useGetQueueInfoByName = () => {
...
+    try {
+      const response = await authedRequest;
+      return response;
+    } catch (error) {
+      history.push(`/pageNotFound/queueName=${queueName}`);
+      return rejectWithValue({ message: `Queue ${queueName} does not exist, try again...` });
     }
-  );
-
-  return getQueueInfoByName;
```
navaneeth-spotnana, line 32: let's revert back to 
```
const getQueueInfoByName = createAsyncThunk(...) { ... };
return getQueueInfoByName;
```

Sorry about this, I will disable this rule on sonarqube

simplq/src/store/asyncActions/setQueueStatus.js
```diff
@@ -23,8 +23,6 @@ const useSetQueueStatus = () => {
     }
     return response;
   });
-
-  return setQueueStatus;
```
navaneeth-spotnana, line 27: Same as above

simplq/src/store/asyncActions/getQueueInfo.js
maaverik: > @maaverik he is using lambda without braces, so the expression becomes the return value:

Ah, my bad, didn't pay attention to the braces.

simplq/.eslintrc.json
```diff
@@ -9,7 +9,7 @@
   "settings": {
     "import/resolver": {
       "node": {
-        "paths": ["src"]
+        "moduleDirectory": ["src", "node_modules"]
       }
```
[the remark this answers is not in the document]
skShekhar: I just tested the code with `paths` and it works fine. I'm not sure why I was getting that warning. I'll revert this change.

simplq/src/store/asyncActions/getQueueInfoByName.js
skShekhar: Reverted the file

simplq/src/store/asyncActions/setQueueStatus.js
skShekhar: Reverted the file

daltonfury42 (approved):
Thanks a lot for the PR!

Now that you have the project setup, would you like to take a stab at a feature implementation? This is a good candidate https://github.com/SimplQ/simplQ-frontend/issues/425

maaverik (approved):

skShekhar:
@daltonfury42 I can work on the max queue feature. I'm new to open source, so I wanted to know where can I post queries for any clarifications related to the feature?

daltonfury42:
For any clarifications needed before you start, you can post comments on the issue thread #425.

Once you have some code ready, you can raise a draft PR and we can also discuss on the PR, if needed.

PR #600, opened 2021-03-29
Add plus to mobile number

navaneeth-spotnana:
The '+' was missing in the phone number data we are sending to backend. Needed to add it for SMS.


simplq/src/components/pages/Join/JoinForm.jsx
```diff
@@ -61,7 +60,7 @@ export function JoinQueueForm({ joinQueueHandler, buttonText }) {
     }
   }
 
-  const handleClick = () => {
+  const onSubmit = () => {
```
navaneeth-spotnana, line 63: `onSubmit` is more descriptive than `handleClick`

maaverik (approved):

PR #603, opened 2021-04-10
Add configurable maximum queue capacity

skShekhar:
This PR implements the configurable maximum queue capacity feature(#425 )
- Moved out `Modal` to the common directory
- Added `disabled` prop in `Button`
- Created `QueueSettings` component which has the side panel option and the settings modal

Screenshots
Q1
Q2
Q3

### DIFF
--- a/simplq/src/components/common/Modal/Modal.jsx
+++ b/simplq/src/components/common/Modal/Modal.jsx
@@ -1,0 +1,34 @@
+import React from 'react';
+import Modal from '@material-ui/core/Modal';
+import { makeStyles } from '@material-ui/core/styles';
+
+const style = makeStyles((theme) => ({
+  root: {
+    overflow: 'scroll',
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+  },
+  content: {
+    position: 'absolute',
+    top: '20%',
+    backgroundColor: '#fff',
+    borderRadius: '25px',
+    outline: 'none',
+    width: 400,
+    padding: 10,
+    [theme.breakpoints.down('sm')]: {
+      width: '95%',
+      margin: 10,
+    },
+  },
+}));
+
+export default ({ open, onClose, children }) => {
+  const classes = style();
+  return (
+    <Modal className={classes.root} open={open} onClose={onClose}>
+      <div className={classes.content}>{children}</div>
+    </Modal>
+  );
+};
--- a/simplq/src/components/common/Modal/Modal.jsx
+++ b/simplq/src/components/common/Modal/Modal.jsx
@@ -1,34 +1,9 @@
 import React from 'react';
 import Modal from '@material-ui/core/Modal';
-import { makeStyles } from '@material-ui/core/styles';
+import styles from './Modal.module.scss';
 
-const style = makeStyles((theme) => ({
-  root: {
-    overflow: 'scroll',
-    display: 'flex',
-    alignItems: 'center',
-    justifyContent: 'center',
-  },
-  content: {
-    position: 'absolute',
-    top: '20%',
-    backgroundColor: '#fff',
-    borderRadius: '25px',
-    outline: 'none',
-    width: 400,
-    padding: 10,
-    [theme.breakpoints.down('sm')]: {
-      width: '95%',
-      margin: 10,
-    },
-  },
-}));
-
-export default ({ open, onClose, children }) => {
-  const classes = style();
-  return (
-    <Modal className={classes.root} open={open} onClose={onClose}>
-      <div className={classes.content}>{children}</div>
-    </Modal>
-  );
-};
+export default ({ open, onClose, children }) => (
+  <Modal className={styles['root']} open={open} onClose={onClose}>
+    <div className={styles['content']}>{children}</div>
+  </Modal>
+);
--- a/simplq/src/components/common/Modal/Modal.module.scss
+++ b/simplq/src/components/common/Modal/Modal.module.scss
@@ -11,12 +11,12 @@
   position: absolute;
   top: 20%;
   background-color: #fff;
-  border-radius: 25px;
+  border-radius: 1.5rem;
   outline: none;
   width: 400px;
-  padding: 10px;
+  padding: 0.6rem;
   @media (max-width: 960px) {
     width: 95%;
-    margin: 10px;
+    margin: 0.6rem;
   }
 }
--- a/simplq/src/components/common/Modal/Modal.module.scss
+++ b/simplq/src/components/common/Modal/Modal.module.scss
@@ -1,0 +1,22 @@
+@import 'components/common.scss';
+
+.root {
+  overflow: scroll;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.content {
+  position: absolute;
+  top: 20%;
+  background-color: #fff;
+  border-radius: 25px;
+  outline: none;
+  width: 400px;
+  padding: 10px;
+  @media (max-width: 960px) {
+    width: 95%;
+    margin: 10px;
+  }
+}
--- a/simplq/src/components/common/Modal/Modal.test.jsx
+++ b/simplq/src/components/common/Modal/Modal.test.jsx
@@ -1,0 +1,15 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import Modal from '.';
+
+describe('Modal', () => {
+  it('should render the modal if it is open', () => {
+    const { getByText } = render(<Modal open>Hello</Modal>);
+    getByText('Hello');
+  });
+
+  it('should not render the modal if it is not open', () => {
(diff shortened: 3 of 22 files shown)


skShekhar:
Had a few questions - 
a. What should be the css for the disabled button
b. Which api should I use to get the current max queue capacity
c. After clicking save, should I close the dialog

simplq/src/components/common/Modal/Modal.jsx
```diff
@@ -0,0 +1,34 @@
...
+    outline: 'none',
+    width: 400,
+    padding: 10,
+    [theme.breakpoints.down('sm')]: {
+      width: '95%',
+      margin: 10,
+    },
+  },
+}));
+
```
maaverik, line 26: Could you move the styles for the content class to SCSS? Since we're using SCSS throughout the project, adding different forms of styling at different places unless really required would set a bad precedent. Ideally, we would want to use SCSS for the root class too, but we can keep makeStyles for root since material-ui is known to have some issues with CSS.

simplq/src/components/common/Modal/Modal.test.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import Modal from '.';
+
+describe('Modal', () => {
+  it('should render the modal if it is open', () => {
+    const { getByText } = render(<Modal open>Hello</Modal>);
+    getByText('Hello');
+  });
+
```
maaverik, line 10: Kudos for adding tests! It's about time we think of incorporating unit tests at least some basic integration tests.

simplq/src/components/pages/Admin/QueueSettings.jsx
```diff
@@ -0,0 +1,87 @@
...
+  title: {
+    display: 'flex',
+    justifyContent: 'center',
+    textDecoration: 'underline',
+  },
+  sizeInput: {
+    marginBottom: theme.spacing(3),
+  },
+}));
+
```
maaverik, line 32: Could you move some of these to SCSS too?

simplq/src/components/pages/Admin/QueueSettings.jsx
```diff
@@ -0,0 +1,87 @@
...
+            value={size}
+            type="number"
+            onChange={handleSizeChange}
+            error={isInvalidSize()}
+            helperText={isInvalidSize() && `Enter a number between 1 and ${MAX_SIZE}`}
+            autoFocus
+          />
+          <div className={classes.actionContainer}>
+            <Button icon={<PrintIcon />} onClick={handleSave} disabled={isInvalidSize()}>
+              Save
```
maaverik, line 77: Could you use the SaveIcon here instead? 

`import SaveIcon from '@material-ui/icons/Save';`

maaverik (changes requested):
Thanks for the PR

maaverik:
> a. What should be the css for the disabled button
The grayed out style looks fine to me.

> b. Which api should I use to get the current max queue capacity
@daltonfury42 

> c. After clicking save, should I close the dialog
That sounds good.

After getting the current capacity, we can also include it in the queue information section.

daltonfury42:
@skShekhar I have just now added two fields to the `GET queue/<queue-id>` endpoint. Sample response:

```
{
   "queueId":"754ae188-f392-4a6d-b5c4-ba39cdff9567",
   "queueName":"ExampleQueue",
   ...
   "maxQueueCapacity":10000,
   "slotsLeft":10000,
   "tokens":[
      ...
   ]
}
```

So you can now define a new `selectMaxQueueCapacity` selector [here](https://github.com/SimplQ/simplQ-frontend/blob/master/simplq/src/store/selectedQueue/selectedQueueSlice.js#L41) and use it. 

While you are at it, can you also add a new extra reducer `updateQueueSettings.fulfilled` that updates the max queue capacity limit?

Thanks

daltonfury42:
> After getting the current capacity, we can also include it in the queue information section.

The [queue info slice](https://github.com/SimplQ/simplQ-frontend/blob/master/simplq/src/store/queueInfo/queueInfoSlice.js) would also similarly have the two new fields, so we can now show them in the QueueInfo component also.

simplq/src/components/common/Modal/Modal.jsx
skShekhar: Moved to scss

simplq/src/components/pages/Admin/QueueSettings.jsx
skShekhar: > > c. After clicking save, should I close the dialog
> 
> That sounds good.
Right now, I'm closing the dialog when user hits save. Is there a way to pass `toggleModal` as a callback to the action so that it will be called only in case of success?

skShekhar:
> > c. After clicking save, should I close the dialog
> 
> That sounds good.

Right now, I'm closing the dialog when user hits save. Is there a way to pass toggleModal as a callback to the action so that it will be called only in case of success?

skShekhar:
> > After getting the current capacity, we can also include it in the queue information section.
> 
> The [queue info slice](https://github.com/SimplQ/simplQ-frontend/blob/master/simplq/src/store/queueInfo/queueInfoSlice.js) would also similarly have the two new fields, so we can now show them in the QueueInfo component also.
I will take this up in a separate PR

navaneeth-spotnana:
> Right now, I'm closing the dialog when user hits save. Is there a way to pass toggleModal as a callback to the action so that it will be called only in case of success?

Hmm.. one way out I can think of is to move `isModalOpen` into an `adminEditForm` slice, and define actions to open and close the modal. What do you think?

simplq/src/components/common/Modal/Modal.module.scss
```diff
@@ -0,0 +1,22 @@
...
+  top: 20%;
+  background-color: #fff;
+  border-radius: 25px;
+  outline: none;
+  width: 400px;
+  padding: 10px;
+  @media (max-width: 960px) {
+    width: 95%;
+    margin: 10px;
+  }
```
maaverik, line 21: One more tiny thing, I missed this in my last review, could you convert the sizing to use rem instead of px. We're using rem just about everywhere else, so might as well keep things consistent.

simplq/src/components/pages/Admin/QueueSettings.jsx
```diff
@@ -0,0 +1,74 @@
...
+import HighlightOffIcon from '@material-ui/icons/HighlightOff';
+import SidePanelItem from 'components/common/SidePanel/SidePanelItem';
+import { useUpdateQueueSettings } from 'store/asyncActions';
+import { useDispatch, useSelector } from 'react-redux';
+import Grid from '@material-ui/core/Grid';
+import SaveIcon from '@material-ui/icons/Save';
+import { selectMaxQueueCapacity } from 'store/selectedQueue';
+import Button from '../../common/Button';
+import InputField from '../../common/InputField';
+import styles from './QueueSettings.module.scss';
```
maaverik, line 13: Instead of long relative imports, you can use 

`import InputField from 'components/common/InputField';`

We've defined the base path to use in jsconfig.json

maaverik (commented):
Just need a few more tiny changes.

simplq/src/components/common/Modal/Modal.module.scss
skShekhar: Fixed

skShekhar:
> Hmm.. one way out I can think of is to move `isModalOpen` into a new `adminEditForm` slice and define actions to open and close the modal. What do you think?

I'm checking for the api response and if there are no errors, dialog is getting closed. This felt simpler than creating a new slice

maaverik (approved):
Great job @skShekhar. We'll merge after our next prod push. By then, @daltonfury42 can also review.

daltonfury42 (approved):

simplq/src/store/selectedQueue/selectedQueueSlice.js
```diff
@@ -29,6 +35,11 @@ const selectedQueueSlice = createSlice({
       state.status = action.payload.status;
       return state;
     },
+    // update queue settings
+    [updateQueueSettings.fulfilled]: (state, action) => {
+      state.maxQueueCapacity = action.payload.maxQueueCapacity;
```
daltonfury42, line 40: Let's set this only if `action.payload.maxQueueCapacity` is undefined?

Thinking out lot, just for extra safety?

skShekhar: Do you mean update the state only if `state.maxQueueCapacity` is undefined OR don't update the state if `action.payload.maxQueueCapacity` is undefined?

daltonfury42:
Actually, ignore. I think backend is supposed to guarantee this field

daltonfury42:
Thanks @skShekhar.

PR #605, opened 2021-04-13
Reverted intentional bugs and removed popup notifcation 

maaverik:
Sentry seems to work fine. The errors don't get logged in the dashboard because we exceeded our quota. ([ref](https://docs.sentry.io/product/accounts/quotas/): "If you’ve exceeded your quota threshold, the server will respond with a 429 HTTP status code.")
Screenshot from 2021-04-13 13-42-10

### DIFF
--- a/simplq/src/components/ErrorHandler.jsx
+++ b/simplq/src/components/ErrorHandler.jsx
@@ -1,6 +1,4 @@
 import React from 'react';
-import { setErrorPopupMessage } from 'store/appSlice';
-import { store } from 'store';
 import * as Sentry from '@sentry/react';
 import PageNotFound from './pages/PageNotFound';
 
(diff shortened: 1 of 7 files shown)


simplq/src/components/ErrorHandler.jsx
```diff
@@ -23,7 +21,6 @@ export class ErrorBoundary extends React.Component {
   componentDidCatch(error, errorInfo) {
     // You can also log the error to an error reporting service
     // logErrorToMyService(error, errorInfo);
-    store.dispatch(setErrorPopupMessage('An error occured. Please try again'));
```
daltonfury42, line 26: Should we be keeping it, or not? 🤔 

What's your points for removing it?

maaverik: I figured it wasn't needed since we were redirecting to the 404 page anyway. It was also never showing up when I tested it, so that'll take some debugging. Do you think we should keep it?

daltonfury42:
Yes, then we should definitely remove it. If something is not getting used,
it’s dead code, should be removed.

daltonfury42 (approved):

PR #613, opened 2021-04-17
Fix for back button not working on 404 page

ShimantaBhuyan:
On clicking join queue button, we are routing to **JoinPage** which dispatches the **getQueueInfoByName** action in its useEffect hook. 
If it fails there, we are redirecting to 404 page. 

The issue now occurs, if we click on back button, we are _going back to JoinPage_ and which _again dispatches the action_ on useEffect. This then _redirects back to 404 page_ and so on, **cyclically**.

I have added a check in the useEffect of JoinPage, to check if the **history action is POP**, which **if true will redirect to home page**, and only if not, will the get queue action be dispatched. 
This might not be the correct way to go about it, but I wasn't able to find a way to get the previous location and do checks on that, without having to store it, which in this case seemed kind of an overkill for me.


maaverik (approved):
Thanks for the PR!

PR #619, opened 2021-04-19
Add queue capacity in queue information section

skShekhar:
This PR adds the queue capacity in queue info section as discussed in #603 .
size

### DIFF
--- a/simplq/src/components/common/QueueInfo/QueueInfo.jsx
+++ b/simplq/src/components/common/QueueInfo/QueueInfo.jsx
@@ -6,10 +6,13 @@ import { useDispatch, useSelector } from 'react-redux';
 import { selectTokens, selectMaxQueueCapacity } from 'store/selectedQueue';
 import styles from './QueueInfo.module.scss';
 
-const DetailRow = ({ title, value, large }) => (
+const DetailRow = ({ title, value, large, valueId }) => (
   <div className={styles['detail-row']}>
     <span className={styles['detail-name']}>{title}</span>
-    <span className={`${styles['detail-value']} ${large ? styles['large-value'] : ''}`}>
+    <span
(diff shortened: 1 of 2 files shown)


simplq/src/components/common/QueueInfo/QueueInfo.jsx
```diff
@@ -40,6 +42,7 @@ export default ({ queueId }) => {
   return (
     <div className={styles['detail']}>
       <DetailRow title="Queue status:" value={status} />
+      <DetailRow title="Queue capacity:" value={queueCapacity} large />
```
daltonfury42, line 45: I think it makes sense to show the slots available to the user than the queue capacity. What do you think?

skShekhar: I don't have a strong opinion here, but I'll prefer total over remaining.

daltonfury42 (approved):
Thank you, for the PR.

simplq/src/components/common/QueueInfo/QueueInfo.jsx
daltonfury42: Yeah, no strong opinion from my side either, thought that number of slots left is a better metric to show the user, as it might interest him more. But should be fine. 👍

maaverik: Just to weigh in here, I also share @daltonfury42's opinion about remaining slots making being more useful to users.

daltonfury42: @skShekhar Let's change to "Available Slots"

skShekhar: Updated to available slots

daltonfury42: Thanks.

PR #620, opened 2021-04-20
Reverting #613 and silencing a warning 

maaverik:
Reverting #613 for now since it might make it difficult for other contributors to work on top of the master branch.


simplq/src/components/common/CreateJoinForm/CreateJoinForm.jsx
```diff
@@ -46,7 +46,7 @@ const CreateJoinForm = (props) => {
       <div className={styles['input-box']}>
         <InputField
           placeholder="Queue Name"
-          value={textFieldValue}
+          value={textFieldValue || ''}
```
maaverik, line 49: There was a warning popping up in the CreateJoinForm component on the browser console. This is to silence it according to [this](https://stackoverflow.com/questions/47012169/a-component-is-changing-an-uncontrolled-input-of-type-text-to-be-controlled-erro) answer.

daltonfury42 (approved):

Issue #621, opened 2021-04-23
UX and Flow Improvements - Draft List


daltonfury42:
1. My Queues doesn't stand out and people will miss it.
2. Scan the QR code takes you to the join page. Join page should have an option to enter the number. If the person is already in the queue, take to the token status page for him. (Self join should be disabled by default, configurable)
3. QR should be present already generated on the admin page than hidden behind a button.

maaverik:
For the india-fights-back branch:

Are you hospital or vaccine center? Start by createing queue:  -> Are you a hospital or vaccine center? Start by creating a queue:

No more crowd at vaccine centers. -> No more crowds at vaccine centers.

Ask them to wait outside without crowding in front of the vaccine . -> Ask them to wait outside without crowding in front of the vaccine centre

Token number can be send via SMS -> Token number can be sent via SMS

Join a queue without crowd -> Join a queue without crowding

When you react the vaccine center, give your mobile number to get into the queue. -> When you reach the vaccine center, give your mobile number to get into the queue.

Waiting for the staff to call you when your turn in near -> Wait for the staff to call you when your turn is near

Check your token status in queue and get notified when your turn is up -> Check your token status in the queue and get notified when your turn is up

Make free pricing say Rs instead of $

sajmalyousef:
Two thoughts:

1.  I may join multiple queues without going to the center . How to address this?
2.  Since this is entirely free. Identity verification might be an issue. Cowin app does aadhar verification. Can the vaccine center allow/reject a person to the queue based on some list they might have?

daltonfury42:
(1) is a real problem. Been thinking about it, and also from a general crowd control perspective. So I am thinking of keeping self-join disabled by default. In the proposed new flow, the queue manager would scan generate the QR code and put it all around the vaccination centre. 

**Flow 1: Default**
1. The queue manager adds the person manually to the queue by asking for his name and contact number. 
2. He would tell the person to wait outside, and to check the status via QR or link send my SMS (SMS will be enabled manually for accounts of covid vaccination centres and hospitals enrolled with us).
3. When the user scans the QRs, it will ask for his mobile number and take him to the token status page.

**Flow 2: Self Join enabled**.
If the admin wants to let people join own their own via the queue link / QR, it will be a configurable queue setting. If admin has enabled "Self Join", upon scanning the QR, it will ask for mobile number as usual. If the mobile number is already in the queue, he will be taken to the token status page, else he will be asked his name, and he can join.

[...]

PR #624, opened 2021-04-24
Get token by phone number and queue id

daltonfury42:
As discussed in https://github.com/SimplQ/simplQ-frontend/issues/621 

Corresponding backend change has to go in together: https://github.com/SimplQ/simplQ-backend/pull/130

Follow up task is to call add the flag to enable self join from queue settings.


simplq/src/components/common/PhoneInput/index.jsx
```diff
@@ -0,0 +1,48 @@
+import React, { useState } from 'react';
```
daltonfury42, line 1: Moved PhoneInput into common for reuse

PR #627, opened 2021-04-24
(bugfix) - Available slots is undefined in Ready to Join page

skShekhar:
**BUG** - Available slots is showing `NaN` since `selectMaxQueueCapacity` is returning undefined in the Ready to Join page.

slots

**FIX** - Replace `selectMaxQueueCapacity` with `selectQueueInfo` to get queue capacity.


daltonfury42 (approved):
Thanks, got very busy, approving after a quick look

PR #629, opened 2021-04-24
Modified admin page to display QR code when empty

maaverik:
Added the QR code to be displayed on the admin page if the queue is empty. The modal is set to open on page refresh, so it'll show up once the queue is created for the first time.

Screenshot from 2021-04-24 18-05-44


daltonfury42 (approved):
1. Scan this QR code it scan QR code to see status
2. No need of automatically opening of modal.

PR #632, opened 2021-04-25
Add chat popup

daltonfury42:


maaverik (approved):

maaverik:
The way it is right now, this shows up on all pages including the status and join pages. We should maybe restrict this to the home and admin pages.

maaverik:
Actually, it's fine. No problems with showing this everywhere.

daltonfury42:
Right, and even more places. I’ll fix this weekend probably

daltonfury42:
I think it should not pop up on mobile, as it takes up the entire space and
is a bit intrusive.

But Pavan and Sajmal says that’s fine. Maybe I feel so because my screen
size is very small.

It’s not at it’s best state currently, I’ll need to iterate more.

PR #633, opened 2021-05-02
'Admin Page Tutorial' improvements

mav1n:
- Step 3: 'Print QR Code' added in Admin Page Tour
- Added done icon as next step icon in the last step of the tour.

### DIFF
--- a/simplq/src/components/common/Popup/QrCode.jsx
+++ b/simplq/src/components/common/Popup/QrCode.jsx
@@ -70,8 +70,11 @@ export const QrCode = ({ queueName, tourTag, handleModalClose }) => {
         queueName={queueName}
         ref={componentPrintRef}
       />
-      <div className={styles['actionContainer']}>
-        <StandardButton tourTag={tourTag} onClick={handlePrint} icon={<PrintIcon />}>
+      <div
+        reactour-selector={tourTag}
+        className={handleModalClose ? styles['actionContainer'] : null}
+      >
+        <StandardButton onClick={handlePrint} icon={<PrintIcon />}>
           Print
         </StandardButton>
         <CloseButton handleModalCloseHandler={handleModalClose} />
--- a/simplq/src/components/common/Popup/QrCode.jsx
+++ b/simplq/src/components/common/Popup/QrCode.jsx
@@ -26,7 +26,7 @@ const ComponentToPrint = forwardRef(({ style, url, queueName }, ref) => {
   );
 });
 
-export const QrCode = ({ queueName, handleModalClose }) => {
+export const QrCode = ({ queueName, tourTag, handleModalClose }) => {
   const componentPrintRef = useRef();
 
   const handlePrint = useReactToPrint({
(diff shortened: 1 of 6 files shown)


simplq/src/components/common/Tour/Tour.jsx
```diff
@@ -24,9 +26,9 @@ export default (props) => {
...
         leftArrow.childNodes[0].style.color = 'grey';
         rightArrow.childNodes[0].style.color = 'white';
-      } else if (stepNumber === props.toursteps.length - 1) {
+      } else {
         leftArrow.childNodes[0].style.color = 'white';
-        rightArrow.childNodes[0].style.color = 'grey';
+        rightArrow.childNodes[0].style.color = 'white';
       }
     } else {
       Sentry.captureMessage('left-arrow or right-arrow selectors of reatTour package not found');
```
daltonfury42, line 34: Let us remove this entire function. Looks hacky, and the default behaviour looks good to me.

daltonfury42 (approved):
Thanks for the PR!

maaverik (approved):
Thanks for the PR. Please see if you can modify it according to @daltonfury42's comments.

simplq/src/components/common/Tour/Tour.jsx
mav1n: image

In the default behaviour, the active 'next' icon is more dull than the disabled 'prev' icon. Shouldn't it be the opposite?

simplq/src/components/common/Popup/QrCode.jsx
```diff
@@ -71,7 +71,7 @@ export const QrCode = ({ queueName, handleModalClose }) => {
         ref={componentPrintRef}
       />
       <div className={styles['actionContainer']}>
-        <StandardButton onClick={handlePrint} icon={<PrintIcon />}>
+        <StandardButton tourTag={tourTag} onClick={handlePrint} icon={<PrintIcon />}>
```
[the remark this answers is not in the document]
mav1n: It should work. I'll check.

simplq/src/components/common/Tour/Tour.jsx
maaverik: Yes, it should be the opposite. In that case, this function will have to stay unless there's a better possible implementation.

daltonfury42: Yeah, I tried locally. 

Right now it's a hacky implementation, and if you go back from the last tutorial, the arrow highlighting is not correct. I would prefer the fix the root cause than fix a bug on top of a hacky implementation.

@maaverik can we remove this function and fix the color as a separate PR?

simplq/src/components/common/Popup/QrCode.jsx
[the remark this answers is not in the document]
mav1n: Updated this in the PR.

simplq/src/components/common/Tour/Tour.jsx
mav1n: I improved the implementation and fixed the arrow highlighting in the same PR.

daltonfury42: @mav1n has handled this elegantly in this PR itself. :)

Issue #403, opened 2020-10-31
Admin Page for deleted queues


daltonfury42:
If someone deletes their queue and goes back using the browser back button to the admin page again, they see the old admin page as it was before. Can we at least show a message in the main admin container saying 'Your queue was deleted' or something like that?

_Originally posted by @maaverik in https://github.com/SimplQ/simplQ-frontend/pull/399#issuecomment-719898266_

How should the admin page look like in case it's deleted? For now, the easiest thing to go is to render a 404 page.

js-sapphire:
Hey @daltonfury42, 
I have tried to handle deleted queues in the Admin Page. Does this solution work ?

daltonfury42:
Hmm. @js-sapphire's PR https://github.com/SimplQ/simplQ-frontend/pull/634 got merged, but there is some issue. Steps to reproduce:

1. Go to home page and create a queue
2. Open the queue in two separate tabs.
3. Delete the queue from one tab.

The expected behaviour after #634 should take the user to the home page with the popup message, but this in not happening.
  
![image](https://user-images.githubusercontent.com/11256376/118362663-11a69980-b5ae-11eb-8bed-c985ec351c8d.png)

I've left some comments on the PR post merge, that could be the reason why it's not happening.

js-sapphire:
I will check this and get back

js-sapphire:
Made code changes in #638

maaverik:
> Hmm. @js-sapphire's PR #634 got merged, but there is some issue. Steps to reproduce:
> 
>     1. Go to home page and create a queue
> 
>     2. Open the queue in two separate tabs.
> 
>     3. Delete the queue from one tab.
> 

@daltonfury42 regarding this issue you pointed out, I suspect it's due to the serviceWorker caching an old copy of the site. I was able to reproduce it in dev.simplq.me, but not on localhost when I used the master branch or incognito dev.simplq.me. On localhost, the behaviour was as expected, the queue was being deleted in the first tab, and after a few seconds, the second tab also followed suit. Could you see if you can consistently reproduce the issue with and without incognito mode? If you can't reproduce it in incognito, I think we should unregister the service worker, it's causing more harm than help.

daltonfury42:
Good catch! It was the serviceWorker caching an old copy of the site.

Serviceworker is something I guess we just have to be aware of. I think let's keep it enabled.

PR #634, opened 2021-05-07
[FEATURE] Handle deleted queues in Admin Page

js-sapphire:
Attempt to solve #403

### DIFF
--- a/simplq/src/components/pages/Admin/AdminPage.jsx
+++ b/simplq/src/components/pages/Admin/AdminPage.jsx
@@ -6,18 +6,44 @@ import { useDispatch, useSelector } from 'react-redux';
 import { selectQueueName } from 'store/selectedQueue';
 import Ribbon from 'components/common/Ribbon';
 import Tour from 'components/common/Tour';
-import { useGetSelectedQueue } from 'store/asyncActions';
+import { useGetSelectedQueue, useGetQueueInfo } from 'store/asyncActions';
+import { selectQueueInfo } from 'store/queueInfo';
 import HeaderSection from './AdminHeaderSection';
 import TokenList from './TokenList';
 import styles from './admin.module.scss';
 import SidePanel from './AdminSidePanel';
 import getToursteps from './getTourSteps';
+import PageNotFound from '../PageNotFound';
 
 const TIMEOUT = 10000;
 let timeoutId;
 
-export default (props) => {
+const isQueueDeleteStatusUnknown = (queueInfo) => !queueInfo || Object.keys(queueInfo).length === 0;
+const isQueueDeleted = (queueInfo) => queueInfo.status === 'DELETED';
+
+const AdminPage = (props) => {
   const queueId = props.match.params.queueId;
+  const queueInfo = useSelector(selectQueueInfo);
+  const dispatch = useDispatch();
+  const getQueueInfo = useCallback(useGetQueueInfo(), []);
+
+  useEffect(() => {
+    dispatch(getQueueInfo({ queueId }));
+  }, [dispatch, queueId, getQueueInfo]);
+
+  if (isQueueDeleteStatusUnknown(queueInfo)) {
+    return <></>;
+  }
+
+  if (isQueueDeleted(queueInfo)) {
+    return <PageNotFound />;
+  }
+
+  return <AdminPageView queueId={queueId} />;
+};
+
+const AdminPageView = (props) => {
+  const { queueId } = props;
   const queueName = useSelector(selectQueueName);
   const description = queueName && 'Ready to share';
   const dispatch = useDispatch();
@@ -62,3 +88,5 @@ export default (props) => {
     </div>
   );
 };
+
+export default AdminPage;
--- a/simplq/src/components/pages/Admin/AdminPage.jsx
+++ b/simplq/src/components/pages/Admin/AdminPage.jsx
@@ -8,38 +8,39 @@ import Ribbon from 'components/common/Ribbon';
 import Tour from 'components/common/Tour';
 import { useGetSelectedQueue, useGetQueueInfo } from 'store/asyncActions';
 import { selectQueueInfo } from 'store/queueInfo';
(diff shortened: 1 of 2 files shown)


maaverik (approved):
Thanks for the PR. I didn't notice any problems on a quick look, this works fine. Later on, we could try showing a page that specifically says something like "This queue has been deleted".

simplq/src/components/pages/Admin/AdminPage.jsx
```diff
@@ -6,18 +6,44 @@ import { useDispatch, useSelector } from 'react-redux';
...
+  const dispatch = useDispatch();
+  const getQueueInfo = useCallback(useGetQueueInfo(), []);
+
+  useEffect(() => {
+    dispatch(getQueueInfo({ queueId }));
+  }, [dispatch, queueId, getQueueInfo]);
+
+  if (isQueueDeleteStatusUnknown(queueInfo)) {
+    return <></>;
+  }
```
daltonfury42, line 36: Why did we have to do this? Curious to know.

Can we avoid it?

simplq/src/components/pages/Admin/AdminPage.jsx
```diff
@@ -6,18 +6,44 @@ import { useDispatch, useSelector } from 'react-redux';
...
+
+  if (isQueueDeleted(queueInfo)) {
+    return <PageNotFound />;
+  }
```
daltonfury42, line 40: in this case, let's use [setErrorPopupMessage](https://github.com/SimplQ/simplQ-frontend/blob/master/simplq/src/store/appSlice.js#L17) to show user a good message and then take the user to the home page using `useHistory`.

It might be better UX than showing the 404 page. What do you think?

cc @maaverik

simplq/src/store/queueInfo/queueInfoSlice.js
```diff
@@ -13,6 +13,13 @@ const queueInfoSlice = createSlice({
     [getQueueInfoByName.fulfilled]: (state, action) => {
       return action.payload;
     },
+    [deleteQueue.fulfilled]: (state, action) => {
+      const { queueId } = action.payload;
+      if (state.queueId === queueId) {
+        return {};
```
daltonfury42, line 19: let us just set the status to deleted inside the if condition

daltonfury42 (changes requested):

simplq/src/components/pages/Admin/AdminPage.jsx
maaverik: True. Explicitly saying that the queue is deleted is more meaningful. For this, just a popup would be as good as a new page.

simplq/src/components/pages/Admin/AdminPage.jsx
js-sapphire: By the time this useSelector hook fetches queueInfo and figures out about the status of the queue, without this the component ends up rendering AdminPageView. I wanted to create a loading state, an error state and a normal state. Hence this code-block.

simplq/src/components/pages/Admin/AdminPage.jsx
js-sapphire: Sure. Will change this.

simplq/src/components/pages/Admin/AdminPage.jsx
daltonfury42: I'm thinking lets continue with normal flow unless we explicitly know that the queue had been deleted.

As discussed, here the only change would be to show the user message and simultaneously take the user to the home page.

simplq/src/store/queueInfo/queueInfoSlice.js
js-sapphire: Done

simplq/src/components/pages/Admin/AdminPage.jsx
js-sapphire: Done

simplq/src/components/pages/Admin/AdminPage.jsx
js-sapphire: Sure. Removed loading state.

daltonfury42 (approved):
Thanks, nicely done!

simplq/src/components/pages/Admin/AdminPage.jsx
```diff
@@ -16,8 +19,31 @@ import getToursteps from './getTourSteps';
...
+const AdminPage = (props) => {
   const queueId = props.match.params.queueId;
+  const queueInfo = useSelector(selectQueueInfo);
+  const dispatch = useDispatch();
+  const history = useHistory();
+  const getQueueInfo = useCallback(useGetQueueInfo(), []);
+
+  useEffect(() => {
+    dispatch(getQueueInfo({ queueId }));
+  }, [dispatch, queueId, getQueueInfo]);
```
daltonfury42, line 33: We don't need to do this, I think getQueueInfo is already periodically called.

simplq/src/components/pages/Admin/AdminPage.jsx
```diff
@@ -16,8 +19,31 @@ import getToursteps from './getTourSteps';
...
+
+  useEffect(() => {
+    if (queueInfo && queueId === queueInfo?.queueId && isQueueDeleted(queueInfo)) {
+      dispatch(setErrorPopupMessage('This queue is deleted.'));
+      history.push('/');
+    }
+  }, [dispatch, history, queueId, queueInfo]);
```
daltonfury42, line 40: We don't need this useEffect, I think. Can we have the if block directly over here?

simplq/src/components/pages/Admin/AdminPage.jsx
js-sapphire: I did it for a different reason. Removing it in #638. 
I thought not just back button, but if any user lands on that url ( queue/queueId ) and Redux store does not know about this queue, I'd make sure we fetch it from server and then check.

PR #637, opened 2021-05-09
Replace history on trying to join invalid queue

### LINKED ISSUE #610: Back button does not work on 404 page.
**Describe the bug**
After trying to join a queue which doesn't exist, pressing the back button on mobile doesn't take you back to the previous website.

**To Reproduce**
1. Go to homepage.
2. Enter a random name of some queue that does not exist.
3. Click on join queue.
4. You'll be presented with 404 page. Pressing back button does not work.

**Expected behavior**
User should be taken back to the home page

maaverik:
Fixes #610. I came across history.replace() while reading an article, didn't know it existed. I haven't noticed any side-effects while testing, so we should be good with this.

### DIFF
--- a/simplq/src/store/asyncActions/getQueueInfoByName.js
+++ b/simplq/src/store/asyncActions/getQueueInfoByName.js
@@ -23,7 +23,7 @@ const useGetQueueInfoByName = () => {
         const response = await authedRequest;
         return response;
       } catch (error) {
-        history.push(`/pageNotFound/queueName=${queueName}`);
+        history.replace(`/pageNotFound/queueName=${queueName}`);
         return rejectWithValue({ message: `Queue ${queueName} does not exist, try again...` });
       }
     }


daltonfury42 (approved):
I lost the context, what was the problem and how does `replace` help?

maaverik:
The problem was that the back button wouldn't work if we reach the 404 page while trying to join an invalid queue. The reason for that was that we redirect to the 404 page from within the join page in case of invalid queues. So when a user clicks the back button, they would go back to the join page with the old selected queue and be stuck in 404 hell. 

Instead of history.push() which adds a page to the history stack while retaining the previous one, the replace() method replaces the top entry of the stack. So in this case, when a user clicks the back button, they go back to the homepage instead of the join page where they get stuck.

daltonfury42:
Nice.

PR #645, opened 2021-08-24
Changes in UI of dialogue box in QueueSettings and copy queue link menu

assassinscreed2:
Hey,
I have made some UI changes in the dialogue box which appears when clicked on the queuesetting button
Previously when clicked on the QueueSettings button it just appears instantly which was not looking that good but now it appears with a small transition effect (fadeIn and fadeout) which is giving a smooth opening and looks nice
and 
I have also implemented a divider in the menu which opens from a side button of the Copy Queue button,
previously It appears a little too plane and now due to dividers it looks nice and separated,
Please have a look 
Thank you

### DIFF
--- a/simplq/src/components/pages/Admin/QueueSettings.jsx
+++ b/simplq/src/components/pages/Admin/QueueSettings.jsx
@@ -1,11 +1,10 @@
 import React, { useState, useEffect } from 'react';
 import SettingsIcon from '@material-ui/icons/Settings';
-import Modal from 'components/common/Modal';
+import { Dialog, Grid } from '@material-ui/core';
 import HighlightOffIcon from '@material-ui/icons/HighlightOff';
 import SidePanelItem from 'components/common/SidePanel/SidePanelItem';
 import { useUpdateQueueSettings } from 'store/asyncActions';
 import { useDispatch, useSelector } from 'react-redux';
-import Grid from '@material-ui/core/Grid';
 import SaveIcon from '@material-ui/icons/Save';
 import { selectMaxQueueCapacity, selectIsSelfJoinAllowed } from 'store/selectedQueue';
 import Button from 'components/common/Button';
@@ -64,7 +63,7 @@ export default ({ queueId }) => {
   return (
     <>
       <SidePanelItem Icon={SettingsIcon} title="Queue Settings" onClick={toggleModal} />
-      <Modal open={isModalOpen}>
+      <Dialog PaperProps={{ style: { width: '27%', borderRadius: 30 } }} open={isModalOpen}>
         <Grid container className={styles['modal-content']} direction="column">
           <h2 className={styles['title']}>Settings</h2>
           <InputField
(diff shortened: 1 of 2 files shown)


simplq/src/components/pages/Admin/QueueSettings.jsx
```diff
@@ -64,7 +63,7 @@ export default ({ queueId }) => {
   return (
     <>
       <SidePanelItem Icon={SettingsIcon} title="Queue Settings" onClick={toggleModal} />
-      <Modal open={isModalOpen}>
+      <Dialog PaperProps={{ style: { width: '27%', borderRadius: 30 } }} open={isModalOpen}>
```
daltonfury42, line 66: Let's make this change [here](https://github.com/SimplQ/simplQ-frontend/blob/master/simplq/src/components/common/Modal/Modal.jsx).

daltonfury42 (commented):
Thanks, I love attention to detail. :)

simplq/src/components/pages/Admin/QueueSettings.jsx
assassinscreed2: I think now we don't need the Modal component, we can replace all the Modal with Dialog components and it will do the all work that the Modal component is doing and more

assassinscreed2:
I have now made the changes in the Modal component, Please have a look , Thank you

simplq/src/components/pages/Admin/QueueSettings.jsx
assassinscreed2: I have now made the changes in the Modal component, Please have a look, Thank you

simplq/src/components/common/Modal/Modal.jsx
```diff
@@ -1,9 +1,8 @@
 import React from 'react';
-import Modal from '@material-ui/core/Modal';
-import styles from './Modal.module.scss';
+import { Dialog } from '@material-ui/core';
 
 export default ({ open, onClose, children }) => (
-  <Modal className={styles['root']} open={open} onClose={onClose}>
-    <div className={styles['content']}>{children}</div>
-  </Modal>
+  <Dialog PaperProps={{ style: { width: '27%', borderRadius: 30 } }} open={open} onClose={onClose}>
```
daltonfury42, line 5: Is it possible to define the CSS in `./Modal.module.scss` and use it here? We want to define styles separately in .scss files wherever possible.

daltonfury42 (commented):
Thanks, it's good to keep as much as components behind a thin wrapper so that we can change things at one place. The PR is mostly good to merge, one final comment.

simplq/src/components/common/Modal/Modal.jsx
assassinscreed2: Yes it is, Actually the main purpose of *.module.scss is that we can define CSS module inside SCSS file which makes it different from a normal *.scss files

assassinscreed2: Yeah Actually it is possible, i will modify it now

assassinscreed2:
I have now defined the CSS for dialog inside SCSS file only. Can you please see it now, Thank You

Issue #601, opened 2021-03-31
Implement Queue History


daltonfury42:
Currently if a token is removed, it's gone for ever from the user's view. We should present the history of deleted tokens in the queue history section on the right side panel.

![image](https://user-images.githubusercontent.com/11256376/113090312-1d5d1d00-9207-11eb-9c1c-79b9ec7d739e.png)

Design: https://xd.adobe.com/view/ad1db074-03bf-45b1-537b-98d9d524ec82-db2c/screen/fe137c74-d5e2-4fe1-af85-10877cc560b8/

For now the log can show just added and removed events.

daltonfury42:
Backend call to `GET /queue/<queue-id>`now returns two lists: the existing list of active tokens, and a separate list of removedTokens, with the deletion time stamp. (https://github.com/SimplQ/simplQ-backend/pull/154)

Mogakamo:
How is supposed to be implemented @daltonfury42

daltonfury42:
The UI design is in the link above, and to get the required details, you can take a look at the response of `GET /queue/<queue-id>` call. Let me know if you have any specific questions while implementing.

anjali-001:
Hey @daltonfury42! I would like to work on this issue.

Soumit915:
@daltonfury42 we need to store a list of objects, that will consists of all the events or history of that particular queue in the backend. 

I want to contribute. Can you please guide me, if we can store this info in the backend? Then, I can make a Queue history as you have mentioned above

anjali-001:
Hey @daltonfury42  I have made a PR(#646 ) that fixes this issue. I made a call to `/queue/<queue-id>' route  to get the list of added and removed tokens, however the removed tokens have the tokenCreationTimestamp, so I made appropriate changes to display the queueHistory list. The details of the changes are mentioned in the PR.

daltonfury42:
There seems to be duplication of efforts here. 

I agree with @Soumit915's point [here](https://github.com/SimplQ/simplQ-backend/issues/126#issuecomment-903574236) that it's better to have a dedicated queue history endpoint that would give a list of events that then we can display. 

One way is to merge @anjali-001 's PR, and then @Soumit915 can work on refactoring it to consume the new endpoint, which you can implement in the meantime. 

@Soumit915 What do you think?

Soumit915:
> There seems to be duplication of efforts here.
> 
[...]

PR #646, opened 2021-08-24
Queue history Implemented

anjali-001:
This fixed issue #601 
This PR has the following things:
- [x] Implementation of Queue History which lists down the added and the removed tokens using the design file given in the issue as a reference.  
- [x] Shows the last two changes and opens up a dropdown on clicking show more and shows last six changes with option to navigate to next or previous changes in a set of 6.

Here are the changes that I made:
Screenshot from 2021-08-24 19-57-07
Screenshot from 2021-08-24 19-57-42

Note:
- Made use of the `extraReducers` to get the removed and added tokens list by making a backend call to `GET /queue/<queue-id>`
- As attached in the image below, in the removed tokens, I am not getting the removed token timestamp and thus the tracking of the removed token is not time precise, though it works without any issue if the page is not refreshed as I am maintaining a `queueHistory` which contains the added and removed tokens sorted according to the time in the local state of the component.

Screenshot from 2021-08-24 20-03-13


simplq/src/store/selectedQueue/selectedQueueSlice.js
```diff
@@ -48,6 +48,8 @@ export default selectedQueueSlice.reducer;
 
 export const selectQueueName = (state) => state.selectedQueue.queueName;
 
+export const selectRemoveTokens = (state) => state.selectedQueue.removedTokens;
```
daltonfury42, line 51: selectRemovedTokens here and other places

simplq/src/components/pages/QueueHistory/AddedInQueue.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import AddIcon from '@material-ui/icons/Add';
+import styles from '../../common/SidePanel/SidePanel.module.scss';
+import itemStyles from './QueueHistory.module.scss';
+
+export default ({ name, tokenNumber, creationTime, tokenCreationTimestamp }) => {
```
daltonfury42, line 6: I feel it would be better have a single HistoryEvent item that can take as parameter if it's a "ADD" or a "REMOVE", what do you think?
```
tokenName, tokenNumber, timestamp, action
```

simplq/src/components/pages/Admin/QueueHistory.jsx
```diff
@@ -1,13 +1,177 @@
...
   return (
     <SidePanelItem
       Icon={HistoryIcon}
       title="Queue History"
       description="History of events in the queue"
-    />
+      expandable
+    >
+      {showMore
+        ? sortedData.slice(count * 6, (count + 1) * 6).map((item) => {
```
daltonfury42, line 104: Let's define 6 on top as a `const MAX_ITEMS`

simplq/src/components/pages/Admin/QueueHistory.jsx
```diff
@@ -1,13 +1,177 @@
...
+            if (item.tokenId === token.tokenId) {
+              isPresent = true;
+            }
+          });
+
+          return !isPresent;
+        });
+      }
+      if (removed) {
+        lastRemoved = await removed.filter((item) => {
```
daltonfury42, line 40: why did we need an await here?

simplq/src/components/pages/Admin/QueueHistory.jsx
```diff
@@ -1,13 +1,177 @@
...
+          let isPresent = false;
+          tokens.forEach((token) => {
+            if (item.tokenId === token.tokenId) {
+              isPresent = true;
+            }
+          });
+
+          return !isPresent;
+        });
+      }
```
daltonfury42, line 38: What's this code doing?

daltonfury42 (commented):
I've left some comments and some questions to understand some parts a bit more.

simplq/src/store/selectedQueue/selectedQueueSlice.js
anjali-001: As per the screenshot that I have attached, we can see that in removedTokens List we are not getting the timestamp of the new removed item, for this I am maintaining the queueHistory and moreover if we refresh we don't have anything in the queueHistory so we can't just add the first item that was newly added  which we could have done in all other cases.

anjali-001: Resolved in latest Comment

simplq/src/components/pages/QueueHistory/AddedInQueue.jsx
anjali-001: Made this into a single file. Resolved in the latest PR

simplq/src/components/pages/Admin/QueueHistory.jsx
anjali-001: Defined this as MAX_ITEMS as asked

simplq/src/components/pages/Admin/QueueHistory.jsx
anjali-001: As per the screenshot that I have attached, we can see that in removedTokens List we are not getting the timestamp of the new removed item, for this I am maintaining the queueHistory and moreover if we refresh we don't have anything in the queueHistory so we can't just add the first item that was newly added  which we could have done in all other cases.

simplq/src/components/pages/Admin/QueueHistory.jsx
anjali-001: I was using a different way to solve this thing and later on I changed it into the present code but forgot to remove await. I have removed it now!

daltonfury42 (approved):
Overall looks good. I would have asked to move the logic inside the useEffect outside the component somehow, but that's fine as #647 would be changing it anyway.

maaverik (approved):

PR #650, opened 2021-09-08
returned expression instead of assigning to temporary variable response

divyansh21jain:


maaverik:
Hi @divyansh21jain, thanks for the PR. This is purely my opinion, but I prefer assigning the response to a temporary variable before returning a tad bit more readable than directly returning.  @daltonfury42 if you think directly returning the expression is better, we can go that route.

daltonfury42:
Yeah, I think it depends on context. I am not having a strong argument
against it, so I’ll say keep the variable. Readability wins.

maaverik:
Closing since we decided to leave it the old way. @divyansh21jain thanks for trying, feel free to take up any other issue that interests you.

PR #651, opened 2021-09-08
Renamed AdminheaderSection file to HeaderSection

divyansh21jain:


maaverik:
The AdminHeaderSection renaming is not required anymore, I think you might have seen an outdated code smell referenced by SonarCube. The component has the same name as the [file](https://github.com/SimplQ/simplQ-frontend/blob/master/simplq/src/components/pages/Admin/AdminHeaderSection.jsx) right now. Also, just a heads-up, while renaming files like this, you would need to remove the original file and change all the imports that reference it, otherwise, this new file is just a copy that will not be used anywhere.

Thanks for trying anyway. Do feel free to take a look at other open issues to see if something interests you.

PR #647, opened 2021-08-26
Solved ScrollTo in Window Issue #636, Implemented the Code for History Queue #601 and Integrated QR code scanning Issue #471

Soumit915:
This fixed issue #636 , provides frontend support for #601 and integrated QR code scanning issue #471 .
The PR consists of: 

- [x] Previously, #636 occured, as Safari doesn't supports the API, Window.scrollTo(option-json). As a result, I, at first, did handle the Safari cases seperately. However, later I implemented an all in one solution and I think it will work. As, now I am passing 2 params for Window.scrollTo, and it won't be an issue as it is supported in Safari. Please confirm me, if the issue gets resolved.
- [x] Provided frontend support for a responsive Queue History #601 . Do need some backend support, with the API format mentioned in [Backend Support](https://github.com/SimplQ/simplQ-backend/issues/126). For now, I have implemented the Queue History for a static data, the dataset is present in `src/data` for your reference. Once the api-route is created from the backend, we can just link the data-fetcher to the api-route instead of the dataset.
- [x] Integrated QR code scanning issue #471 . Created a button in home page that will redirect to `/scanQr` route and you can scan any QR code which will then redirect you to target url. Haven't put any condition or any constraint on what the url can be, but you can please update me with the constraints if any, so that I can apply and update it accordingly. The scanning can be done with even your computer or laptop if you have a valid camera device. For mac users or safari customers, where scanning is an issue, the app takes captures images of the qr-code and then processes it. However, the users need not to worry anything about it as the process for a mac-user is same as that of any user using any other device.
- 
image
Glimpse of the frontend ui, implemented.

- [x] I have also re-added the pause queue, functionality which was previously muted.

PR #652, opened 2021-09-08
Queue History issue #601 completed and continuation of PR #647

Soumit915:
This now solves issue #601 with real-time api.

This is basically a continuation of PR #647 .

I have now redesigned the ui for Queue-History. Implemented it with real-time api from the backend. After each add and remove token, the queue-history gets updated within 10 seconds. The user can also update the queue-history by clicking the Refresh Status button.

Please leave your reviews and merge if you like it.


maaverik (approved):
Works as expected on a quick try. Great job!

simplq/src/store/selectedQueueHistory/selectedQueueHistorySlice.js
```diff
@@ -0,0 +1,24 @@
...
+} from 'store/asyncActions';
+
+const selectedQueueHistorySlice = createSlice({
+  name: 'selectedQueueHistory',
+  initialState: {
+    events: [],
+  },
+  reducers: {},
+  extraReducers: {
+    // handle fulfilled request
```
daltonfury42, line 15: As a follow up PR, can you also subscribe to `joinQueue.fulfilled` and `deleteToken.fulfilled` events and update the events state by adding events to it?

Ref [here](https://github.com/SimplQ/simplQ-frontend/blob/master/simplq/src/store/selectedQueue/selectedQueueSlice.js#L24)

daltonfury42 (approved):

simplq/src/components/common/QueueHistory/HistoryRow.js
```diff
@@ -36,19 +37,19 @@ export default (props) => {
   //for medium-sized to larger-sized devices, function returns a larger icon
   const getIcon = () => {
     const fontSize = "180%";
-    return props.data.action === 'ADDED'
+    return props.data.eventType === 'TOKEN_ADDED'
       ? <AddIcon  style={{fontSize: fontSize}} />
-      : props.data.action === 'REMOVED'
+      : props.data.eventType === 'TOKEN_REMOVED'
```
daltonfury42, line 42: Post merge comment: I would have preferred a if else here for readability over a complex ternary if operator. 😄

daltonfury42: image

Soumit915: @daltonfury42 , you are right I also do prefer if-else. Idk why the hell I did a ternary there. Will surely change it to an if-else, and update the ui also of the `/scanQr` route in my next PR

simplq/src/store/selectedQueueHistory/selectedQueueHistorySlice.js
Soumit915: @daltonfury for that, the post response should be updated I think. Like currently the action payload for deleteToken:fulfilled request, consists of the following json, which doesn't consist of the expected eventTimestamp. I think the backend delete response should be updated so that to update as you are wanting

{
    "tokenId": "b3abb7b2-7a5e-4513-a511-796f48fe8c1a",
    "queueName": "Myq1",
    "tokenStatus": "REMOVED"
}

Soumit915: @daltonfury42 , I can update things for joinQueue.fultilled request, by manipulating the incomming action.payload and taking the relevent fields. But one need to update the delete api response. Can you please see that and make the changes in the backend, so that I can update changes in the frontend

Issue #654, opened 2021-09-26
Increase font size for reading purposes


rahulshawdev:
**Is your feature request related to a problem? Please describe.**
The font size of this section is too small. People with eye problems will have a hard time reading.

**Describe the solution you'd like**
Maybe increase the font size.

**Additional context**
![simplq-increase](https://user-images.githubusercontent.com/82671048/134802989-244aaf22-e4d9-4847-b579-c72d040b17f9.png)

maaverik:
Hi @rahulshawdev, that's a fair ask. Do you want to give this item a try? I think [this](https://github.com/SimplQ/simplQ-frontend/blob/a9ff4f2e97b3c9e3743d22924d1583b0fe3a3632/simplq/src/components/pages/Home/Home.module.scss#L171) CSS class is the one that styles this part.

rahulshawdev:
yeah sure :)

daltonfury42:
Sure, the first PR would be accepted, though. 🏎️ 

Lets also increase to the next level and not too much.

PR #664, opened 2021-10-05
fix(home): increase font-size #654

iqbalmuhyiddin:
hi, I just fix this issue. hope can help.


maaverik:
You would also want to increase font-size for the detailed-description CSS class, that's the one which styles the text below the three headings.

iqbalmuhyiddin:
ah i see. okay. is that okay if i just remove the `font-size:small`? so it will follow the browser default font-size?

PR #671, opened 2021-10-08
On join page list user's tokens created in the past

daltonfury42:
On join page list user's tokens created in the past, so that he can get easily to the token status page.

No more need to remember the token link. You can close tabs and come back to the join link.

https://user-images.githubusercontent.com/11256376/136613362-dff4d9ad-dda3-437b-8728-25bcb9f7cb73.mov

### DIFF
--- a/simplq/src/api/requestFactory/index.js
+++ b/simplq/src/api/requestFactory/index.js
@@ -28,4 +28,11 @@ export {
   updateQueueSettings,
 } from './queue';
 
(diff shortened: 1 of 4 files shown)


daltonfury42:
> Is the flow that, first, all token for a user are fetched and then, they're filtered based on the current queue name before being displayed?

Yes. Generally the queue link is shared via whatsapp etc. So the person can go to the link again and get back to his token.

I want to try and improve the styling a bit, if I get in the mood to do so.

Issue #673, opened 2021-10-09
Fix crash: ReferenceError: Intl is not defined


daltonfury42:
[This line](https://github.com/SimplQ/simplQ-frontend/blob/master/simplq/src/components/common/Footer/Footer.jsx#L10) at times crashes with stacktrace:
```
ReferenceError: Intl is not defined
  at c (components/common/Footer/Footer.jsx:10:29)
  at Module.312 (index.jsx:30:10)
  at l (/j/Mars-9Oct-Saturday:1:2308)
  at Object.254 (None:2:46)
  at l (/j/Mars-9Oct-Saturday:1:2308)
...
(14 additional frame(s) were not displayed)
```

![image](https://user-images.githubusercontent.com/11256376/136666217-87c9a7d6-4169-4d26-9c5a-405d8238fe6c.png)

This was reported on a "MiuiBrowser", so looks like the Intl module is not present in some non-standard browsers: 

To fix this, add a check to see in Intl is `undefined`. If so, the function `dayOfWeek` can return the string `"day"`.

indremak:
I would like to work on this issue, could you assign it to me?

daltonfury42:
Sure, go ahead. We don’t usually assign issues, but the first complete PR
will be accepted.

PR #674, opened 2021-10-09
Add check if Intl is defined, set default day value

indremak:
Fix for #673 Fix crash: ReferenceError: Intl is not defined


daltonfury42:
I intentionally introduced the typo to make the error happen, as it was trying to load module that is not present in the runtime.

indremak:
added checks for `Intl` and `DateTimeFormat` types, could you review?

PR #686, opened 2021-12-31
Package version upgrades

maaverik:
npm install was failing for me on a new OS with higher node version (v16), so I thought I'd try to update packages to get it to install and fix vulnerabilities along the way. npm reported >100 vulnerabilities with existing versions. 

This PR need not be merged unless we absolutely have to upgrade versions. It's only partially been tested, so bugs probably remain. I successfully updated most modules, and got the vulnerabilities reduced to 37. I had to uninstall the qr-scanner module, but it wasn’t working properly anyway, some console errors still remain, and eslint reports some false positive lint errors. I'll get back to fixing this sometime, but just documenting what I did here in case the same thing has to ever be done again someday.

Steps so far:

1. Updated packages - https://nodejs.dev/learn/update-all-the-nodejs-dependencies-to-their-latest-version
Removed react-qr-scanner package due to incompatibility with version of react-scripts
2. Running dev server failed due to a gulp dependency error - fixed with https://stackoverflow.com/a/60921145/10281310
3. Module not found error came up for react-refresh - fixed by changing versions of react-refresh-webpack-plugin as described in https://github.com/storybookjs/storybook/issues/17049 and installing react-refresh package
4. react-typeform-embed export not found - fixed by replacing existing component with Popup https://www.npmjs.com/package/react-typeform-embed - still work in progress to fix an error
5. React-router v6 has different interfaces for routing - made changes according to https://stackoverflow.com/a/69849271/10281310, https://stackoverflow.com/questions/62861269/attempted-import-error-usehistory-is-not-exported-from-react-router-dom, https://stackoverflow.com/a/69796015/10281310
6. babel-eslint not found in .eslintrc - replaced with @babel/eslint-parser
7. eslint throwing errors in browser instead of warnings in console, fixed according to https://stackoverflow.com/a/64630234/10281310


daltonfury42:
Ah. M1 pains!

